PR #441, opened 2014-02-06
Add support for group by in select

GabrielDelepine:
I made a jsfiddle : http://jsfiddle.net/yappli/X8era/9/

I used AngularStrap for 2 days only. Thanks for this project and sorry if I didn't respect some rules angularStrap's developpers should know. I'm ok to learn it !


LennyLip:
Any news? And why only "group by"?
May be we need select with tree support?
Like a:
http://a5hik.github.io/angular-multi-select-tree/

PR #976, opened 2014-08-14
tooltip.js - prefixEvent as data attribute

arnoudsietsema:
Added 'prefixEvent' to the list of supported data attribute options


vmlf01:
@mgcrea The show/hide events trigger with the provider instance as argument, so maybe we could add a new attribute/option, like a key, that would be put on a provider attribute so the event handler can test which specific instance triggered the event? This could be applied to all tooltip derived directives.

What do you think?

mgcrea:
@vmlf01 idea landed in c0a1fb8709ff65ee6d8f6c3d44d938d39e7055cd.

Issue #602, opened 2014-03-25
overriding $typeahead.$onKeyDown


masrawi:
I am trying to customize the the typeahead select behavior  by adding my own template to the directive and mainly overriding the on-click value

```
ng-click="onSelect(match.value)"
```

which works as expected for click selections but the problem is when the user  is pressing the enter on a selection. digging into the code the $typeahead.$onKeyDown is invoked. how is it possible to override this?

PR #1075, opened 2014-09-16
allow overriding of events in typeahead

masrawi:
the events of typeahead are now extracted to their own service allowing the easy decoration of
them.

https://github.com/mgcrea/angular-strap/issues/602


christophertrudel:
@mgcrea any plans on supporting the use of decorators on your services? I am also interested in this enhancement. I guess for now I can fork and apply something like the suggested fix here but it would really be helpful to expose key functions on your services.

PR #1111, opened 2014-09-29
refactor(typeahead): better updating and selectMode

ruifortes:
change updating mecanism from model watcher to parser
this provide better control of whats actually gets to model
"selectMode" uses options "value" attribute and changes model only upon valid selection
When not in "selectMode" model value is synchronized with input string and upon selection changes to options "label" attribute
"formater" to update viewValue use ~~new "$parseOptions.parseViewValue"~~ $parseOptions.displayValue method. (it was since added in another PR)

matches are only updated on input change (via parser) so simply focusing a prestine control would just create an empty list

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -381,6 +381,7 @@ gulp.task('views:pages', function() {
 //
 var jshint = require('gulp-jshint');
 var stylish = require('jshint-stylish');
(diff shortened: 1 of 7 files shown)


mgcrea:
Replaced by #1310.

PR #1142, opened 2014-10-11
bs checkboxgroup will now work with ng repeat

almondtools:


iwuvjhdva:
Possible to merge?

mgcrea:
@almondtools Could you please remove the built file (dist) and squash your commits into a single one? Thanks!

almondtools:
I am not so familiar with git. Is the commit correct now?

By the way: Is the patch acceptable? I do not mind, if it is reworked by others, if it does not suite into the rest of the code.

mgcrea:
The patch fails against latest 1.3, I think you should use `bs-checkbox` directly when you want to use ng-repeat. Added test related to that.

PR #1163, opened 2014-10-16
Wrap all date/locale function in new $dateUtil service

tomyam1:
Currently the date and time pickers use `$dateParse` for formatting, `dateFilter` for formatting and `$locale` to determine the formats.

This works fine for most applications, but sometimes we want to use another library, e.g. [moment](http://momentjs.com/) to this work for us. This is true especially, if we already use moment for other components. Otherwise, we'd have to keep two sets of locale data, and there still might be some incompatibilities between the two.

This intermediate wrapper is needed to allow this customization. Check for example, [angular-strap with moment](http://plnkr.co/edit/wGVxTg0rwizLLGBeypmO?p=preview) plunker.

Since the wrapper is placed in the `date-parser.js` and in the  `mgcrea.ngStrap.helpers.dateParser` module, this PC is completely backward compatible.

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -1,6 +1,9 @@
 'use strict';
 
-angular.module('mgcrea.ngStrap.datepicker', ['mgcrea.ngStrap.helpers.dateParser', 'mgcrea.ngStrap.tooltip'])
+angular.module('mgcrea.ngStrap.datepicker', [
+  'mgcrea.ngStrap.helpers.dateParser',
+  'mgcrea.ngStrap.helpers.dateFormatter',
+  'mgcrea.ngStrap.tooltip'])
 
   .provider('$datepicker', function() {
 
@@ -20,6 +23,10 @@ angular.module('mgcrea.ngStrap.datepicker', ['mgcrea.ngStrap.helpers.dateParser'
       dateFormat: 'shortDate',
       modelDateFormat: null,
       dayFormat: 'dd',
+      monthFormat: 'MMM',
+      yearFormat: 'yyyy',
+      monthTitleFormat: 'MMMM yyyy',
+      yearTitleFormat: 'yyyy',
       strictFormat: false,
       autoclose: false,
       minDate: -Infinity,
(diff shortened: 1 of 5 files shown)


matthewdfleming:
In case anyone has needed to override $dateParser.. here's how I did it to add the ability to parse multiple date formats when the user types in the value themselves.. (e.g. 02242016, 2-12-2015, 2017-12-21).

I didn't spend a ton of time trying to add this as a feature to strap.. so the number only parsing has a fixed format (instead of doing something like base it on the $locale format stripped of non-numeric characters for instance). The wrapper uses moment (and it's known formats) to do the heavy lifting.

    (function (angular) {
        'use strict';

        angular.module('momentStrap', ['mgcrea.ngStrap.helpers.dateParser'])

            .config(function ($provide) {

                /**
                 * Wraps the existing $dateParser helper used by angular strap
                 */
                $provide.decorator('$dateParser', function ($delegate, moment) {
                    return function (config) {
                        var dateParser = $delegate(config),
                            isValidFn = dateParser.isValid,
                            parseFn = dateParser.parse,
                            getDateForAttributeFn = dateParser.getDateForAttribute,
                            getTimeForAttributeFn = dateParser.getTimeForAttribute,
                            daylightSavingAdjustFn = dateParser.daylightSavingAdjust;

                        // delegate to existing function
                        dateParser.isValid = function (date) {
                            return isValidFn(date);
                        };

                        /**
                         * Allows the parser to be more flexible when the user enters dates
                         *
                         * @param value the date string
                         * @param baseDate used by the default parser
                         * @param format sometimes passed to this function
                         * @returns {*} a date or false when bad
                         */
                        dateParser.parse = function (value, baseDate, format) {
                            format = format || dateParser.$format; // use the configured format if not passed
                            if (isFinite(value)) {
                                // value is a number, attempt to parse using MMddyyyy format
                                format = 'MMddyyyy';
                            }
                            var ret = parseFn(value, baseDate, format);
                            if (!ret && !isFinite(value)) {
                                // try to parse with moment
                                ret = moment(value);
                                ret = ret.isValid() ? ret.toDate() : false;
                            }
                            return ret;
                        };

                        // delegate to existing function
                        dateParser.getDateForAttribute = function (key, value) {
                            return getDateForAttributeFn(key, value);
                        };

                        // delegate to existing function
                        dateParser.getTimeForAttribute = function (key, value) {
                            return getTimeForAttributeFn(key, value);
                        };

                        // delegate to existing function
                        dateParser.daylightSavingAdjust = function (date) {
                            return daylightSavingAdjustFn(date);
                        };

                        return dateParser;
                    };
                });
            });
    })(window.angular);

Issue #1164, opened 2014-10-16
typeahead: trigger does not work


smuellerus:
I am using typeahead, and tried triggering the drop-down with a click / input focus, but it does not work, even when using

```
 data-trigger="focus"
```

(which should be the default)

It only triggers when I start typing.

Any ideas?

PR #1173, opened 2014-10-20
fix(typeahead): show dropdown on focus when minLength is 0

vmlf01:
The first time the input control receives focus and minLegth was 0, the options dropdown was not showing. If you typed a character and then deleted it, then the dropdown would always show on focus. Turns out that the first time, the view value was undefined, so the internal $isVisible value was returning false. Fixed it by defaulting undefined display values to empty string.

Might be the cause for #1164.

This is happening on the docs page with the icons input.


nevenavv:
AFAICS, this has been released with 2.1.6, but I'm still experiencing this issue with 2.1.6

Issue #960, opened 2014-08-07
bs-affix-target ...how does it work?


k-funk:
I can't seem to figure out how bs-affix-target should work. Can you put an example in the docs?

Furthermore, I really think that the css styling should be left up to the user, completely. Certain inline-styles get applied by angular-strap, making it hard to override them. The affix, changing position:fixed; is by nature, kind of a css hack, making the parent node's width/container irrelevant once the node is out of document flow.
Hacking around inline styles is a mess.

The BS docs say "Use the affix plugin via data attributes or manually with your own JavaScript. In both situations, you must provide CSS for the positioning and width of your affixed content." 
... there's a reason they haven't 'solved' the resulting bug, and it's because css (external style sheets) should be the only deciding factor.

PR #1208, opened 2014-10-31
affix: optional inline styles

k-funk:
Reference: #960 

I use the panel from affix on mobile (xs and sm), but don't want it to have an affix-behavior on mobile. The inline styles that are currently applied in .js make it near-impossible to use my own css (there's no room for a fixed sidebar/column).

### DIFF
--- a/src/affix/affix.js
+++ b/src/affix/affix.js
@@ -5,7 +5,8 @@ angular.module('mgcrea.ngStrap.affix', ['mgcrea.ngStrap.helpers.dimensions', 'mg
   .provider('$affix', function() {
 
     var defaults = this.defaults = {
-      offsetTop: 'auto'
+      offsetTop: 'auto',
+      inlineStyles: true
     };
 
     this.$get = function($window, debounce, dimensions) {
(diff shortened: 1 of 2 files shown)


src/affix/affix.js
```diff
@@ -216,9 +227,14 @@ angular.module('mgcrea.ngStrap.affix', ['mgcrea.ngStrap.helpers.dimensions', 'mg
       require: '^?bsAffixTarget',
       link: function postLink(scope, element, attr, affixTarget) {
 
-        var options = {scope: scope, offsetTop: 'auto', target: affixTarget ? affixTarget.$element : angular.element($window)};
-        angular.forEach(['offsetTop', 'offsetBottom', 'offsetParent', 'offsetUnpin'], function(key) {
-          if(angular.isDefined(attr[key])) options[key] = attr[key];
+        var options = {scope: scope, target: affixTarget ? affixTarget.$element : angular.element($window)};
```
k-funk: Removed `offsetTop` and `inlineStyles` from this options object, so that the provider defaults can serve their purpose.
@vmlf01 @mgcrea

mgcrea:
Thanks! Landed in bfd917b.

PR #1210, opened 2014-10-31
feat(typeahead): expose ability to specify comparator

bstruthers:
A comparator can be chained onto the filter in Angular.  This exposes the ability to add a comparator to the filter that controls what shows in the typeahead.


nevenavv:
If options is array of objects, comparator fn gets called for every property of an object. Is there a way that it gets called only for certian property or a whole object (as expected value that should be compared to input)?

PR #1223, opened 2014-11-03
feat(tooltip): add ability to auto position tooltip

bstruthers:
Add ability for tooltip (and any dependent directives) to be auto
positioned.  Placement option has been enhanced to look for `auto`.  If
present, tooltip will try to position itself in viewport.  A preferred
placement can also be specified by appending desired placement to auto.
Tested against Win 7 Chrome, Firefox, IE 9-11.

### DIFF
--- a/src/tooltip/tooltip.js
+++ b/src/tooltip/tooltip.js
@@ -306,22 +306,28 @@ angular.module('mgcrea.ngStrap.tooltip', ['mgcrea.ngStrap.helpers.dimensions'])
 
         $tooltip.$applyPlacement = function() {
           if(!tipElement) return;
+          
+          // Determine if we're doing an auto or normal placement
+          var placement = options.placement,
+              autoToken = /\s?auto?\s?/i,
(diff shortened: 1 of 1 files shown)


smajl:
Eh.. just found out that the auto placement (which is mentioned long time all over the docs) is not working with tooltip.. and probably not with other directives either.. because the feature has not been yet merged. That's kinda misleading...

bstruthers:
@smajl This has been in the source for a while.  Are you mixing this up with the viewport changes in 2.2.0?  

See this plnkr (using 2.1.6) for an example of how auto works.  Since placing the tooltip on the left causes it to be outside the bounds of the example div, it is automatically positioned to the right.  http://plnkr.co/edit/Le3hoBxu5f6QQAVwy4Ba?p=preview

smajl:
@bstruthers Oh, I get it. Vague docs made me think that the position is calculated against viewport, but in fact, it's calculated against placement container. It works then, but it should be more clear.

Issue #1233, opened 2014-11-05
Tabs makes my form $dirty


just-boris:
I have a form divided by tabs, like this

``` html
<form name="form.myform" class="form-horizontal"
      ng-submit="save()">
    <div bs-tabs ng-model="selectedTab">
        <div data-title="Common" bs-pane>
            //some inputs
        </div>
        <div data-title="Extra" bs-pane>
            //some another inputs
        </div>
    </div>
    <div class="pull-right">
            <button type="button" class="btn btn-default" ng-click="close()">Close</button>
            <input type="submit" class="btn btn-default" ng-class="{'btn-success': form.myform.$dirty}" value="Save changes" />
     </div>
</form>
```

I track `form.$dirty` property to show message for user about unsaved changes
Without `ng-model` on tabs all working good, but when I had added it, form remains `$dirty` forever.
What do I do? I think, that is bad idea to bind active tab to model, because it is not actual change which should be saved on server.

joaoborsoi:
I have the same issue with collapse. There is already an open issue at https://github.com/mgcrea/angular-strap/issues/1561 but with no answer. It would be great a similar solution.

PR #1235, opened 2014-11-06
feat(tab): add bsActivePane attribute to bind active tab pane

vmlf01:
Currently the tab directive is using ngModel to bind the active pane index. This causes the tab control to be included in the validation pipeline and when used inside a form, causes the form to be $dirty. (see issue #1233)

This commit deprecates usage of ngModel and adds a replacement bsActivePane attribute with the same functionality, so it should be a matter of replacing ngModel references with bsActivePane.

### DIFF
--- a/src/tab/docs/tab.demo.html
+++ b/src/tab/docs/tab.demo.html
@@ -14,13 +14,13 @@
 $scope.tabs.activeTab = {{ tabs.activeTab }};
   </pre>
   <div class="bs-example" append-source>
-    <!-- ngModel is optional -->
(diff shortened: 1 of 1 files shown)


bstruthers:
@mgcrea & @vmlf01 how long do you plan on supporting ngModel with tab.js?  Is there a plan for removing it?

mgcrea:
@bstruthers I think we should keep supporting it, it could fit some use cases like using a vertical tab as a radio-like form controller. What do you think?

bstruthers:
That works for me.  If we plan on keeping it around, can we remove the console.warn that says it's been deprecated?

PR #1275, opened 2014-11-11
feat(bsRadio): Added support for value being modified by ngValue directive

mdvorak:
Right now, usage of ngValue directive with bsRadio is not supported. This patch observes value attribute instead of evaluating it during linking, therefore reacts for its changes.

This is now supported:
`<input type="radio" bs-radio ng-model="test" ng-value="scopeValue"/>`


mgcrea:
Thanks, landed in 4e0a98c.

PR #1292, opened 2014-11-17
Remove opened datepicker cause error

cgarnier:
When an auther component have removed an opened datepicker, we got error 'off undefined' because $element no more existe.

This error can appears in a ngGrid custom template.


mgcrea:
Closed for inactivity

PR #1310, opened 2014-11-23
refactor(typeahead): major restructuring

ruifortes:
change updating mecanism from model watcher to parser
this provide better control of what actually gets to model
"selectMode" uses ngoptions "value" and changes model only upon valid selection
when not in "selectMode" model value updated with input string and upon selection changes to ngoptions "label"
"formater" to update viewValue use new "$parseOptions.displayValue" method


mgcrea:
Replaced by #1332

ansonkao:
@ruifortes is this PR still being developed? If it's stale, I'd like to submit my own PR for the issue of the cursor moving to the end, as it's a pretty urgent bug for me. Please let me know!

Issue #1011, opened 2014-08-28
Default selected value in typeahead show bad result


Ehres:
Hello,
When i use typeahead who return id and display name, the selected value display id.
The bug is reproduce on this plunker :
http://plnkr.co/edit/BNL4S8NhX3RghQfeXIFL?p=preview

Issue #652, opened 2014-04-08
Typeahead with object as model behaving inconsistently in Firefox


nnjpp:
This is reproducible on the demo page (used Firefox 28.0 on both Mac OS X and Windows 7).

In the icon selection demo, selecting an icon should set the model to the corresponding object:
{"value":"Globe","label":"<i class=\"fa fa-globe\"></i> Globe"}

This sometimes works as it does in all other browsers, but about 50% of the time the model value is set to the view value, "Globe".

Vanitas1440:
Any update on this? @nnjpp 's workaround is working nicely with v2.1.6 / Angular v1.3.13

ranjansatya:
Is there any update on this ? This bug is still there in 2.2.0

MagicIndustries:
Is anyone working on this? I have the same problem - trying to use three typeaheads making them related, IE select an option in typeahead 3 and it will auto-select the relevant options in typeahead1 and typeahead2. It works great unless I have previously typed into one of the typeaheads.

This works:
1) Load page, all typeaheads are empty
2) start typing in typeahead3, select match 
3) typeahead 1 and 2 are auto-selected with the right options.

But if I do this:
1) Load page, all typeaheads are empty
2) start typing in typeahead1, select match.
3) start typing in typeahead3, select match
4) typeahead 2 is auto-selected with the right option, typeahead 1 remains on my original selection.

OR
1) Load page, all typeaheads are empty
2) start typing in typeahead1, backspace until the text box is empty.
3) start typing in typeahead3, select match
4) typeahead 2 is auto-selected with the right option, typeahead 1 remains empty.

Essentially whatever I try to tweak, it works UNLESS a typeahead has previously had something typed into it. If I just load the page and keep changing typeahead3's selection, the rest update fine, but stop when I manually enter text in, even if I remove it afterward.

I'm assuming it's something to do with checking whether there was a previously stored value, but I can't seem to find the right bit of underlying code to fiddle with.

just-boris:
I think that would be better to get rid of storing matched value in the ngModel at all. It is counter-intuitive: if you matched a suggested value, you will get an object, if there is no match, you will get a raw string.

It is better to have `ngModel` with input value all time, but suggested object can be caught in the `$typehead.select` listener. It would make things in the code of typeahead more simpler.

ansonkao:
[...]

PR #1332, opened 2014-12-04
Typeahead refactor select mode and model update

vmlf01:
Added some more tests and fixes on top of @ruifortes PR #1310, but still have some problems when testing in docs pages:
- Selecting item with ENTER key doesn't update ngModel bindings
- Icon input (minLength = 0) doesn't show options dropdown until you type something, then it will behave as expected
- In Firefox, when selecting an icon with the mouse the ngModel will be set to the item, but then pressing tab or clicking outside, the ngModel will be updated again with the input text (there were already some previous issue reports related to this: #652, #1011)
- In Angular 1.2, when using selectMode, the input will be ng-valid even when a value has not been selected (ngModel = undefined), causing 1 test to fail.

This is weird because the tests will pass, but it clearly show the problems using the docs pages. Except for the last one, I wasn't able to make the tests fail to reproduce these issues. Need help to fix them :(

### DIFF
--- a/src/typeahead/docs/typeahead.demo.html
+++ b/src/typeahead/docs/typeahead.demo.html
@@ -40,12 +40,18 @@ $scope.selectedAddress = "{{selectedAddress}}";
         <label><i class="fa fa-home"></i> Address <small>(async via maps.googleapis.com)</small></label>
         <input type="text" class="form-control" ng-model="selectedAddress" data-animation="am-flip-x" ng-options="address.formatted_address as address.formatted_address for address in getAddress($viewValue)" placeholder="Enter address" bs-typeahead>
       </div>
+      <hr>
+      <div class="form-group">
+        <label><i class="fa fa-globe"></i> State <small>(with selectMode="true")</small></label>
+        <input type="text" class="form-control" ng-model="selectedState" ng-options="state for state in states" placeholder="Enter state" data-select-mode="true" bs-typeahead>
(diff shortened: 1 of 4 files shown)


ansonkao:
@vmlf01 is this PR still being developed? If it's stale, I'd like to submit my own PR for the issue of the cursor moving to the end, as it's a pretty urgent bug for me. Please let me know!

vmlf01:
@ansonkao I think it would be best to submit your PR

edhenderson:
@ansonkao Hey, did you submit your own PR for the cursor-to-the-end-problem? LMK, thanks. E.

ansonkao:
@edhenderson I didn't end up tackling it. There is an issue on GitHub explaining how to solve the problem, I have it written down somewhere, if you wanted to try I can dig it up!

edhenderson:
@ansonkao If you have it that would be great. I tried something involving catching the keyUp and keyDown to retro-put the cursor back but when you delete too quickly, the Typeahead event wins the race condition.

ansonkao:
@edhenderson found it. Have a look through these 2 links, hope it helps!

https://github.com/mgcrea/angular-strap/issues/1448
http://stackoverflow.com/questions/26263169/how-to-retain-cursor-position-when-updating-a-knockout-js-observable-in-an-exten?rq=1

edhenderson:
:+1:

PR #1357, opened 2014-12-16
fix(modal): Allow modal element to be available in before event

SimeonC:
Moved the modal.show.before event to occur immediately after the element is created.

I'm not sure if there is anything bad that could happen from this, but I wanted to be able to use the before event to place the modal over the button that triggered the modal.

Tests failed but that seemed to be something unrelated to do with the timePicker.


mgcrea:
Thanks, landed in d6670ff.

PR #1367, opened 2014-12-18
Disable browser auto complete

cgarnier:
The browser auto complete is in conflict with this component.
The issue appears on Firefox: there are two auto complete.


mgcrea:
Thanks, landed in 9a31edc.

PR #1380, opened 2014-12-26
Update select.js to parse iconCheckmark option

arathael:
It was not into the options array of keys to parse. It would be nice to implement a test into the demos.


mgcrea:
Thanks! Landed in 3cbc2b9.

PR #1389, opened 2014-12-30
Added a progress bar directive, with tests and docs

### LINKED ISSUE #379: Progressbar
Is there any plan to implement support for Bootstraps progressbar? It's the only thing I'm missing compared to angular.ui.bootstrap.

DaveWM:
resolves #379 , adds a progress bar directive. Allows you to enable/disable animations and select the type of progress bar (success, warning, etc.).

### DIFF
--- a/src/progressbar/progressbar.js
+++ b/src/progressbar/progressbar.js
@@ -4,7 +4,7 @@ angular.module('mgcrea.ngStrap.progressbar', [])
 .provider('$progressbar', function (){
     var defaults = {
       type: '',
-      animate: function (){ return true;}
+      animate: true
     };
 
     this.$get = function (){
@@ -22,11 +22,11 @@ angular.module('mgcrea.ngStrap.progressbar', [])
       scope:{
         value: '=',
         type: '@',
-        animate: '&'
+        animate: '=?'
       },
       link: function (scope, element, attr){
         scope.type = scope.type || $progressbar.defaults.type;
-        scope.animate = angular.isDefined(scope.animate()) ? scope.animate : $progressbar.defaults.animate;
+        scope.animate = angular.isDefined(scope.animate) ? scope.animate : $progressbar.defaults.animate;
       }
     };
   });
--- a/src/progressbar/progressbar.js
+++ b/src/progressbar/progressbar.js
@@ -3,7 +3,7 @@
 angular.module('mgcrea.ngStrap.progressbar', [])
 .provider('$progressbar', function (){
     var defaults = {
-      barType: '',
+      type: '',
       animate: function (){ return true;}
     };
 
@@ -25,16 +25,8 @@ angular.module('mgcrea.ngStrap.progressbar', [])
         animate: '&'
       },
       link: function (scope, element, attr){
-        scope.type = scope.type || $progressbar.defaults.barType;
+        scope.type = scope.type || $progressbar.defaults.type;
         scope.animate = angular.isDefined(scope.animate()) ? scope.animate : $progressbar.defaults.animate;
-        scope.$watch('type', function (){
-          if(scope.type) {
-            scope.barClass = 'progress-bar-' + scope.type;
-          }
-          else{
-            scope.barClass = null;
-          }
-        });
       }
     };
   });
--- a/src/progressbar/progressbar.js
+++ b/src/progressbar/progressbar.js
@@ -1,0 +1,40 @@
+'use strict';
+
+angular.module('mgcrea.ngStrap.progressbar', [])
+.provider('$progressbar', function (){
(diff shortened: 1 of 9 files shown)


mgcrea:
Thanks for the PR, but I can't review it as you've committed the build files, please remove them and squash your commits into a single one with a proper message format (check the guidelines).

DaveWM:
@mgcrea Done, hadn't seen the contributing guidelines sorry

manikantag:
Does this allows using progress bar along with buttons? ex: kind of https://github.com/remotty/angular-ladda. Not exactly like this module, but will this allows adding progress bar to any other directive too?

Thanks.

DaveWM:
@manikantag The progress bar directive compiles to a div, so it should work within a button yes

manikantag:
oh, good then.

vmlf01:
@mgcrea This looks good, should it be merged?

mgcrea:
@vmlf01, I'd like to merge this, but it feels like it needs some tweaks to be better aligned with the rest of the library (mostly scope handling & var naming). Needs a bit more investigation when I'll have the time.

src/progressbar/progressbar.js
```diff
@@ -0,0 +1,40 @@
...
+      };
+    };
+  })
+  .directive('bsProgressbar', function ($progressbar){
+    return {
+      restrict: 'E',
+      transclude: true,
+      replace: true,
+      templateUrl: 'progressbar/progressbar.tpl.html',
+      scope:{
```
mgcrea: AngularStrap uses `scope: true` accross the project, why use `&` for animate (and `@` for type)?

src/progressbar/progressbar.js
```diff
@@ -0,0 +1,40 @@
+'use strict';
+
+angular.module('mgcrea.ngStrap.progressbar', [])
+.provider('$progressbar', function (){
+    var defaults = {
+      barType: '',
```
mgcrea: should be `type`, simpler is better

src/progressbar/progressbar.js
```diff
@@ -0,0 +1,40 @@
...
+        value: '=',
+        type: '@',
+        animate: '&'
+      },
+      link: function (scope, element, attr){
+        scope.type = scope.type || $progressbar.defaults.barType;
+        scope.animate = angular.isDefined(scope.animate()) ? scope.animate : $progressbar.defaults.animate;
+        scope.$watch('type', function (){
+          if(scope.type) {
+            scope.barClass = 'progress-bar-' + scope.type;
```
mgcrea: barClass does not really needs to live on the scope, it should be concatenated in the `ng-class` inside the template

src/progressbar/progressbar.js
DaveWM: I was using `&` for animate because I didn't know optional 2 way bindings existed, I've updated it to `=?` :). What's the advantage of using `scope:true`?

henry74:
Is this available in the latest build?

vmlf01:
No, it hasn't been merged yet

rzschech:
It would be nice if this change:
- supported min/max not being 0/100 and computes the width percentage based on the min and max.
- supported min-width as per the Bootstrap progress bar examples
- the output {{value}}% span is optional

DaveWM:
@rzschech 
* I did consider allowing the min/max values to be changed, but I think it's simpler to just convert the value to a percentage. Having said that, if you really need this feature I can add it in.
* You can set the min-width via css:
```css
.progress-bar{
    min-width: 2em;
}
```
* The content within the progress bar is already optional. You can add whatever you like in here, like so:
```html
<bs-progressbar value="value">My Content</bs-progressbar>
```

src/progressbar/progressbar.js
mgcrea: In the past, the isolated scopes were buggy, and leaked to other directives (see angular/angular.js/issues/1924). So AngularStrap avoided them since they were mostly decorator directives and this bug broke ngModel bindings.
It is now fixed but the project has not migrated yet, for the sake of consistency, I'd rather keep defaulting to scope:true (leaving the isolated scope definition in comments to ease a future update). At some point (maybe for the next minor), we should migrate all directives at once.

mgcrea:
Regarding the naming, I'm wondering if the naming shouldn't be `bs-progress-bar` & `mgcrea.strap.progressBar`, looks consistent with the official bootstrap naming.

DeividasK:
Can't wait to see this PR merged.

hermansje:
Hopefully this can be merged soon!

billyaraujo:
+1.

JoelParke:
+1

Issue #1390, opened 2014-12-30
Collapse Animation


CarlosCuevas:
The default animation doesn't work unless you include http://mgcrea.github.io/angular-strap/styles/main.min.css

psi-4ward:
my `slide-up-down-ani` class for ng-repeat and collapse

```javascript
app.animation('.slide-up-down-ani', function() {
  function enter(element, done) {
    jQuery(element).css({
      display: 'none'
    }).slideDown(200, done);
  }

  function leave(element, done) {
    jQuery(element).css({
      display: 'block',
      visibility: 'visible'
    }).slideUp(200, done);
  }

  return {
    enter: enter,
    leave: leave,
    addClass: function(el, className, done) {
      if(className !== 'in') return;
      enter(el, done);
    },
    removeClass: function(el, className, done) {
      if(className !== 'in') return;
      leave(el, done);
    }
  };
});
```

sfarfarini:
The documentation misses the `@keyframe` part, because `animation-name` refers to `expand` and `collapse`, which are simple not in place in any standard css.
I extracted them by myself:

```
@keyframes expand {
  from { max-height: 0; }
  to { max-height: 500px; }
}

@keyframes collapse {
  from { max-height: 500px; }
  to { max-height: 0; }
}
```

This is confusing, because not specifying the animation in the custom css seems they are taken from some standard-css that should be included, like from angular-motion or bootstrap, but it-s not.
Please add them to the docs or change the animation to something which does not require custom css.

belfz:
What's the status of this bug? I still can't make the bs-collapse animations working. Any solutions?

vmlf01:
Possible animation implementation, as mentioned above:
http://plnkr.co/edit/fV1RXB5zsCHnuwBR8N9b?p=preview

belfz:
Thank you, that works fine.

PR #1392, opened 2015-01-01
docs(collapse): add custom CSS for panel animation to docs

vmlf01:
As pointed out in #1390, the CSS listed in the docs was outdated. Added the custom CSS used in the docs (extracted from the http://mgcrea.github.io/angular-strap/styles/main.min.css file)

### DIFF
--- a/src/collapse/docs/collapse.demo.html
+++ b/src/collapse/docs/collapse.demo.html
@@ -72,24 +72,54 @@ $scope.panels.activePanel = {{ panels.activePanel }};
 
   <div class="callout callout-info">
     <h4>Custom animations</h4>
-    <p>Pane animation is done with the <code>active</code> class and requires custom CSS.</p>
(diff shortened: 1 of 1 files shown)


mgcrea:
Do you think we should directly link the angular-motion css file use in the docs instead: [collapse.css](https://github.com/mgcrea/angular-motion/blob/master/dist/modules/collapse.css), I'm not sure to want this big blocks of CSS texts in the docs, (I should add tabs for that...)?

vmlf01:
Yes, that is probably best because there would be several blocks of CSS spread throughout the docs. Maybe create some design pattern to make sure they stand out so people notice them and realize they need the CSS?

mgcrea:
I think I'd drop the CSS prefixes (for the sake of vertical space) and add a two-pane system like http://mgcrea.github.io/angular-strap/#/tabs-usage

PR #1393, opened 2015-01-02
All/none buttons margin bug

arathael:
Right margin was not declared.


vmlf01:
Why do you need the margin-right? I don't think there is anything to the right of the buttons.

mgcrea:
Please provide context and a plunker if you want it to be reopened ;-). Thanks.

PR #1404, opened 2015-01-06
docs(plunker): fix angular-motion and bootstrap-additions CSS urls

vmlf01:


mgcrea:
Thanks! Landed in 500ac17.

PR #1406, opened 2015-01-07
Make sure these buttons do not submit the form.

twogood:
Hi,

I noticed that clicking one of the All and None buttons also submitted my form. Setting type="button" (as opposed to default type="submit") fixes this. 

Another solution would be to add event.preventDefault();  to the ng-click handlers.


mgcrea:
Thanks! Landed in c9d454a.

PR #1408, opened 2015-01-07
docs(datepicker): fix startWeek option default value

vmlf01:


mgcrea:
Thanks, landed in 7eb51ff.

PR #1414, opened 2015-01-08
Added autoClose option to select 

demee:
that makes it possible to use with 'click' trigger 
click opens, and then auto-close="true" allows to close it if clicked outside, leveraging tootltip functionality


mgcrea:
Thanks, landed in 4e5d493.

PR #1420, opened 2015-01-09
Add directive to force format on datepicker inputs

mmouterde:
inspirated by https://github.com/AlphaGit/ng-pattern-restrict modified to autocomplete and fit with bsDatePicker dateFormat.
Work with dateFormat pattern EXCEPTED for those that are not mask (one to one characters) :
=> Do not work with these pattern : 'y','MMMM', 'MMM', 'M', 'd', 'EEEE', 'EEE', 'H', 'h', 'm', 's','a','Z','w'
=> work with these pattern : 'yy','yyyy', 'MM', 'dd', 'HH', 'hh', 'mm', 'ss', 'sss','ww'
=> work with localizable formats with the same limits.

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -409,6 +409,156 @@ angular.module('mgcrea.ngStrap.datepicker', [
     };
 
   })
+  
+  .directive('mask', function($locale) {
+		/*
+     *	inspirated by https://github.com/AlphaGit/ng-pattern-restrict modified to autocomplete and fit with bsDatePicker dateFormat.
+		 *  Work with dateFormat pattern EXCEPTED for those that are not mask (one to one characters)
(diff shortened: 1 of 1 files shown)


mgcrea:
This can't be merged as is, it lacks docs and unit tests.
Moreover, angular-strap provides a `$dateParser` service that should be used to check mask validity.
Finally, I'm not really this really sure yet if this should be part of AngularStrap.
Thanks for the work anyway, I'll leave it open to see if there is intereset.

mmouterde:
I agree with you about the leak of doc & test.

This directive is not a matter of validation. It prevents to fill illegal characters and autocomplete separators. Tell me if you change your mind and you decide to integrate a such feature. I probably  complete doc & test.

mgcrea:
@mmouterde I think it sould live in a separate repository have you considered creating a repository for it?

PR #1425, opened 2015-01-11
Feature/add triggering element id to events

vmlf01:
Sorry for so many commits, they are basically the same for each of the directives. This should implement the $id attribute on all directives that trigger events.


mgcrea:
Thanks! Landed ;-).

PR #1426, opened 2015-01-11
fix(select): fix sort option when value is false

### LINKED ISSUE #1421: feat: keeping the order in which items are clicked in multi-select
In the multiselect sample if I click on 'Camera' and then 'Heart', the order of the items in the selectedIcons array is :  'Heart', 'Camera'.

Is there a way to keep the order in which they were clicked ?

vmlf01:
Adds support for parsing the sort option attribute.
Fixes #1421


mgcrea:
Thanks! Landed in 627f3f7.

Issue #1402, opened 2015-01-06
prefixEvent option needed in Typeahead


AdrianMF:
Noticed that prefixEvent is used to help distinguish events in multiple modals, but for some reason it is not allowed when using the Typeahead module. I have multiple typeahead modules on the same page and this option would be helpful.

dok3:
If you add prefixEvent to array in the line 178 of typeahead.js, the attribute "prefix-event" will change the prefixEvent option value: 
```javascript
178  angular.forEach(['prefixEvent', 'placement', 'container', 'delay', 'trigger', 'keyboard', 'html', 'animation', 'template', 'filter', 'limit', 'minLength', 'watchOptions', 'selectMode', 'comparator', 'id'], function(key) {
179      if(angular.isDefined(attr[key])) options[key] = attr[key];
180 });
```

vmlf01:
Check out the new ```id``` option in the latest version, should be easier and more helpful.

dok3:
Probably i'm wrong, but we are catching the emit with the prefixEvent option from here (typeahead.js):
```javascript
82 scope.$emit(options.prefixEvent + '.select', value, index, $typeahead);
```
If you have an easier way to do this, please, explain it to us.

vmlf01:
With the new ```id``` option you only need to register for one event and the handler will be called for all instances. Then in the handler you can test the instance ```id``` that triggered the event. Say you set ```prefixEvent``` to ```my```:

```
scope.$on('my.select', function(evt, value, index, $select) {
  var sourceId = $select.$id;
  ...
});

```

AdrianMF:
The 'id' option is exactly what I was looking for. Thanks for the update.

rtorrero:
Is there any reason why the prefixEvent can't be interacted with in the typeahead options? In the documentation, prefixEvent is not listed as an available option (though listening for $typeahead.select, which is the default value, works)

vmlf01:
Don't know, it's probably just an overlook.
Anyway, I think you probably would use the id option instead from now on, I don't really see a use case for a custom prefix on an instance basis.

PR #1427, opened 2015-01-11
prefixEvent editable by the prefix-event attribute

dok3:
The prefix-event attribute is not changing the prefixEvent option value.


vmlf01:
Related to issue #1402

mgcrea:
Check the new id option. Thanks.

PR #1429, opened 2015-01-12
feat(tooltip): Added support for bootstrap like viewport overflow adjust...

michaelfdeberry:
Added functionality to adjust the tooltip position so that it does not overflow the viewport. 
sample


BrandonLWhite:
I was just about to open an issue for this exact problem.  It seems to me like the tooltip "auto" isn't quite taking into account all of the possibilities that it needs to be.

[Here's a plunker](http://plnkr.co/edit/Ipwhv5c443QmW8JKAaxP?p=preview) to demonstrate the problem.

mgcrea:
Thanks for the PR, I'll check the changes asap ;-).

gaui:
Why not use `window.innerHeight` and `window.innerWidth` instead of the dimensions of the `html` DOM element?

mgcrea:
Thanks for the PR, #1472 was merged and it solves this issue.

PR #1433, opened 2015-01-14
(typeahead) add trimValue feature

ianus1234:
The new option allows to disable the view value of the input field to be trimmed. it might be necessary if searching for several words where the request starts directly after entering a space. this space will then not be removed from the view value.


kkirsche:
I'd recommend fixing the tests to work with your change to increase the likelihood of it being merged. :+1:

vmlf01:
You should make your changes in the source files in the src folder and only commit those. The files in the dist folder are generated by the build process when a new release is made.

ianus1234:
thx for the feedback

PR #1434, opened 2015-01-14
chore(nuget): various nuget related changes

khellang:
Just an FYI; I'm still keeping the NuGet package up to date :smile: It's got 6,199 downloads right now :+1:


mgcrea:
@khellang Thanks!

PR #1439, opened 2015-01-15
(typeahead) add trimValue feature

ianus1234:
The new option allows to disable the view value of the input field to be trimmed. it might be necessary if searching for several words where the request starts directly after entering a space. this space will then not be removed from the view value.


mgcrea:
Could you please squash all your commits into a single one and use a proper commit message?

qqilihq:
Hi guys, what's the state on this? I'm having problems where spaces at the end during entering a value are just stripped off, which is very annoying. Sounds like these changes could solve my issue...

chrisguerrero:
+1

I'm waiting on this to get merged as well

blaind:
+1

jcassee:
+1

rtorrero:
I just tried this, and I'm having trouble using it in combination with delayed model update. Could anyone try ng-model-options="{debounce: 500}" and confirm this? 

Without this, it seems to work correctly.

sank:
+1
Trimming appears when you get a list from the server.
ng-model-options is the same problem.

Issue #1100, opened 2014-09-24
Problem: show date in datepicker with different timezone


deshaser:
I have Date object: 2014-09-02T00:00:00+03:00.
For example if user's local timezone is GMT +4 he will see in datepicker 2014-09-02 01:00 because datepicker converts time to his local time.
Can we set up datapicker to ignore local user's timezone and force it to needed timezone (for example GMT +3)? Or can we pass extarnal filter for the attribute "type-format"?

Issue #885, opened 2014-07-07
Setting a different timezone for the DateTimePicker


xJREB:
If I use a combination of the date and the time picker, it would be really cool to be able to set the timezone programmatically for it to alter the presentation of my date. The underlying numerical value (unixtime in ms) should of course remain untouched. 

I suppose until now there is no easy solution for this (at least I couldn't find one in the docs/code), so consider this as a feature request. :) Is something like this planned? And if not, how would you recommend to handle a case like this?

PR #1449, opened 2015-01-21
feat(datepicker,timepicker): add timezone option

rbebic:
Since Angular's [dateFilter](https://docs.angularjs.org/api/ng/filter/date) already supports UTC timezone. This was a quick solution. I'm not sure if this is the right approach, but it can be extended to support all timezones and it worked for me.

```javascript
angular.extend($datepickerProvider.defaults, {
    dateFormat: 'dd/MM/yyyy',
    startWeek: 1,
    timezone: 'UTC'
});

angular.extend($timepickerProvider.defaults, {
    arrowBehavior: 'picker',
    timeFormat: 'HH:mm',
    length: 1,
    timezone: 'UTC'
});
```

Issues: [#885](https://github.com/mgcrea/angular-strap/issues/885), [#1100](https://github.com/mgcrea/angular-strap/issues/1100)

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -21,6 +21,7 @@ angular.module('mgcrea.ngStrap.datepicker', [
       useNative: false,
       dateType: 'date',
       dateFormat: 'shortDate',
+      timezone: null,
       modelDateFormat: null,
       dayFormat: 'dd',
       monthFormat: 'MMM',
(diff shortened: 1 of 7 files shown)


mgcrea:
Nice! Thanks for the PR, would you mind adding unit tests for the timezone option in both datepicker and timepicker?

rbebic:
Done! However, to calculate clients timezone offset with respect to non UTC timezone is going to involve some kind of table that maps timezone names to their offsets (taking into account DST). But let's wait for Angular to add support for this first, maybe we could reuse something from it :)

mgcrea:
@vmlf01 I'm considering merging this for the next release, any thoughts again it?

@rbebic Thanks for adding the tests, could you please squash your commits into a single one to avoid history pollution?

rbebic:
@mgcrea Thank you! It would be great to have this feature in next release.

I had squashed my commits.

mgcrea:
Thanks! Landed in 9aad910eed818b6e005459e1500ce31d5dde9cfc.

vmlf01:
This sound great! So if we use something like momentjs timezone to parse/format the dates, we could display times in any timezone, right?

AdaskoTheBeAsT:
I hit recently that issue in important business project when we had to allow users to choose dates and times in given timezone  - as a workaround i used additional field which is binded to date and time picker and in which i recalculate date using choosen timezone (and time zone of a user browser) - i used angular-moment as a base solution and reworked it to my needs. Please forgive me for mediocre code quality - I came few months ago to angular from c# world ;)   http://plnkr.co/edit/bLpcXl?p=preview

lkachun9129:
Found a problem with timezone set to 'UTC', the case is as below:
There are two timepickers, A and B. B's min value is bounded by A's value. When the timezone of both are set to 'UTC', the min value comparsion will ignore the timezone setting and still be processed with localized timezone adjustment.

I applied a temp fix by calculating the timezoneOffsetAdjustment before compare with min value.
```javascript
$timepicker.$isDisabled = function (date, index) {
	var dateParser = $dateParser({
		format: options.timeFormat,
		lang: lang
	});
	var selectedTime;
	var dateAdjusted = dateParser.timezoneOffsetAdjust(date, options.timezone, true);
	if (index === 0) {
		selectedTime = dateAdjusted.getTime(), +viewDate.minute * 6e4 + viewDate.second * 1e3;
	} else if (index === 1) {
		selectedTime = dateAdjusted.getTime() + viewDate.hour * 36e5 + viewDate.second * 1e3;
	} else if (index === 2) {
		selectedTime = dateAdjusted.getTime() + viewDate.hour * 36e5 + viewDate.minute * 6e4;
	}
	return selectedTime < options.minTime * 1 || selectedTime > options.maxTime * 1;
};
```

Issue #1447, opened 2015-01-21
Tooltip module not working on IE8


JoshSchreuder:
The tooltip module is no longer working in IE8 following the changes to auto position tooltips (specifically commit @adb68eeba66cddede9afe897c04764afb75440e9)

This also breaks anything that uses tooltip, eg. datepicker / timepicker.

The line that causes issues in IE8 is:
https://github.com/mgcrea/angular-strap/commit/adb68eeba66cddede9afe897c04764afb75440e9#diff-e8665eb306ddd26399e5b80f00dae5bbR468

This goes through Angular's ForEach which fails when it tries to iterate the IHTMLRect object `elRect`.

This line can be confirmed to be responsible by simply returning elPos instead of the concatenated object, which returns behaviour back to pre-auto-position standard.

JoshSchreuder:
@bstruthers any ideas on this one? I'm not sure if it's an issue with Angular or not. My first workaround was to try something like:
```
var pos = {
  top: elPos.top || elRect.top,
  bottom: elPos.bottom || elRect.bottom
  ...
}
```

which is an ugly way of copying the two objects together, but it seemed to work. You'll probably have a better fix in mind.

bstruthers:
@JoshSchreuder What version of Angular are you using?  I'll try to recreate on my machine, so far I've been unsuccessful with getting IE8 to work at all :)

JoshSchreuder:
Whatever the latest Angular 1.2 is, I'll check when I'm at work.

Angular.extend goes into Angular.forEach which is where the issue occurs, with the `for key in obj` loop

vmlf01:
Seems related to this:
https://github.com/angular/angular.js/issues/5201

JoshSchreuder:
@vmlf01 I found that ticket too, but they seemed resigned to the fact it was only happening on emulated IE8 through document standards. 
This is happening to me on actual IE8.

JoshSchreuder:
@bstruthers using Angular 1.2.28. I will try and put together a minimal Plnkr case and post it to angular/angular.js#5201 and see what they say.

JoshSchreuder:
OK, posted a test case to that Angular ticket.
http://plnkr.co/edit/q876cl6Dkr5OowkL2nWO?p=preview

bstruthers:
Here's what I've learned, in IE8 getBoundingClientRect is a read-only object. In other browsers, the object can be manipulated.

[...]

PR #1456, opened 2015-01-22
fix(tooltip): copy properties of bounding rectangle to support IE8

bstruthers:
IE8's TextRectangle object is read-only, causing an exception inside
angular.extend.  Copying the values of the object into a new object
allows the properties to be used inside angular.extend and fixes issue
#1447.

### DIFF
--- a/src/tooltip/tooltip.js
+++ b/src/tooltip/tooltip.js
@@ -476,9 +476,19 @@ angular.module('mgcrea.ngStrap.tooltip', ['mgcrea.ngStrap.helpers.dimensions'])
           var el = $element[0];
 
           var elRect = el.getBoundingClientRect();
+          var rect = {};
+          
+          // IE8 has issues with angular.extend and using elRect directly.
+          // By coping the values of elRect into a new object, we can continue to use extend
+          for (var p in elRect) {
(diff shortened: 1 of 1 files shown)


mgcrea:
Thanks! Landed in 61ee4f9.

JoshSchreuder:
@bstruthers, just had a chance to check this out in v2.2.0, seems like it's still failing because IE8 doesn't support `hasOwnProperty()` on the TextRectangle :frowning:

bstruthers:
My bad, in my test plnkr I did not have the hasOwnProperty check because I was just proving out the read-only theory.  When I added it to AngularStrap, I did so out of habit.  

IE8/9 etc do support hasOwnProperty on the TextRectangle.  The problem is, it always returns false.  Since hasOwnProperty is supposed to be 'frozen' in all browsers, we should be able to safely remove the hasOwnProperty check.

bstruthers:
@mgcrea I submitted a PR with the removal of hasOwnProperty.  If you want me to do something different, let me know.

@JoshSchreuder  This should hopefully address your IE woes.  I verified that the tooltips were positioning correctly in IE8,9,10,11, Firefox, and Chrome.

PR #1459, opened 2015-01-23
fix(bsTooltip): ignore target attribute when parsing options

### LINKED ISSUE #1454: Can't add tooltip on link with target attribute
I want to show hints about external links on my site:
```html
<a href="http://github.com" target="_blank" bs-tooltip data-title="Opens in new tab">Open github</a>
```

It is not works and I see an exception on console for each hover
```
Uncaught TypeError: Cannot read property 'getBoundingClientRect' of undefined 
```

It caused by that tooltip try to find a target from attribute. But it is not setting for it. Maybe it should to read only value from `data-target` not just `target`

Here is reproduction of issue http://plnkr.co/edit/CMj7nX?p=preview

just-boris:
closes #1454

### DIFF
--- a/src/tooltip/tooltip.js
+++ b/src/tooltip/tooltip.js
@@ -619,9 +619,14 @@ angular.module('mgcrea.ngStrap.tooltip', ['mgcrea.ngStrap.helpers.dimensions'])
 
         // Directive options
         var options = {scope: scope};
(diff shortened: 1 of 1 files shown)


mgcrea:
Thanks! Landed in 1a2425c.

PR #1465, opened 2015-01-24
docs(navbar): fix navbar docs home url route match rule

### LINKED ISSUE #738: Navbar active element
**Home** link should not be active when navigating to **Page 1** and **Page 2**.

Instead of:
```
<li data-match-route="/?(navbar)?"><a href="#/">Home</a></li>
```

documentaton should be updated to:
```
<li data-match-route="/$"><a href="#/">Home</a></li>
```

I am not even sure what is `?(navbar)?` supposed to do in this case?

vmlf01:
Closes #738


mgcrea:
Thanks! Landed in 80fda24.

PR #1466, opened 2015-01-24
fix(modal): correctly parse boolean options attribute values

### LINKED ISSUE #1462: modal backdrop still shows despite option being false
In the modal directive, the boolean option data-backdrop="false" still adds a backdrop.
The "true" and "static" options work as expected, but "false" does not.

vmlf01:
Fixes #1462


mgcrea:
Thanks! Landed in 0a44397.

PR #1468, opened 2015-01-25
added main so that jspm knows what to include

capaj:
now you can simply require('angular-strap'), because jspm is expecting this on every package


mgcrea:
Landed in 3af4172.

capaj:
@mgcrea thanks.

Issue #1207, opened 2014-10-30
Support upstream tooltip `viewport` option, respect bounds of the viewport


blah238:
https://github.com/twbs/bootstrap/pull/12328
http://getbootstrap.com/examples/tooltip-viewport/

Would be great to have this in Angular-Strap. Specifically I'm looking to use this for popovers, but it should work for those as well.

blah238:
Tentatively closing this since it is [now in v2.2.0](https://github.com/mgcrea/angular-strap/commit/c9daa6166113358dc044fbe421b8332dd9984c4c).

PR #1472, opened 2015-01-27
feat(tooltip): Add support for Bootstrap's viewport positioning

bstruthers:
Adds support for keeping tooltips and dependent controls inside the
viewport.  Follows the Bootstrap implementation.  This closes issue #1207.

### DIFF
--- a/src/tooltip/test/tooltip.spec.js
+++ b/src/tooltip/test/tooltip.spec.js
@@ -67,55 +67,73 @@ describe('tooltip', function() {
       element: '<a data-animation="am-flip-x" bs-tooltip="tooltip">hover me</a>'
     },
     'options-placement-top': {
-      element: '<a data-placement="top" bs-tooltip="tooltip">hover me</a>'
+      element: '<a data-placement="top" bs-tooltip="tooltip" data-viewport="null">hover me</a>'
     },
     'options-placement-right': {
-      element: '<a data-placement="right" bs-tooltip="tooltip">hover me</a>'
+      element: '<a data-placement="right" bs-tooltip="tooltip" data-viewport="null">hover me</a>'
     },
     'options-placement-bottom': {
-      element: '<a data-placement="bottom" bs-tooltip="tooltip">hover me</a>'
+      element: '<a data-placement="bottom" bs-tooltip="tooltip" data-viewport="null">hover me</a>'
     },
     'options-placement-left': {
-      element: '<a data-placement="left" bs-tooltip="tooltip">hover me</a>'
+      element: '<a data-placement="left" bs-tooltip="tooltip" data-viewport="null">hover me</a>'
     },
     'options-placement-exotic-top-left': {
-      element: '<a data-placement="top-left" bs-tooltip="tooltip">hover me</a>'
+      element: '<a data-placement="top-left" bs-tooltip="tooltip" data-viewport="null">hover me</a>'
     },
     'options-placement-exotic-top-right': {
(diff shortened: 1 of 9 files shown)


Hrafnkellos:
:+1:

gaui:
:+1: We need this. When the container is `body` and we're using bootstrap, the viewport is incorrect.

src/tooltip/test/tooltip.spec.js
```diff
@@ -67,55 +67,73 @@ describe('tooltip', function() {
       element: '<a data-animation="am-flip-x" bs-tooltip="tooltip">hover me</a>'
     },
     'options-placement-top': {
-      element: '<a data-placement="top" bs-tooltip="tooltip">hover me</a>'
+      element: '<a data-placement="top" bs-tooltip="tooltip" data-viewport="null">hover me</a>'
```
mgcrea: Is this really required? Does it means this is a breaking change for any existing implementation?

mgcrea:
Quite a big patch! Similar to #1429.

src/tooltip/test/tooltip.spec.js
bstruthers: @mgcre the bootstrap default is to have the body as the viewport.  If the viewport wasn't null for the standard placements and auto placements, things would shift causing the tests to have to account for it.  I thought it would be easier to test those placements with one less variable.

Hrafnkellos:
i think he is trying to mock what is in bootstraps original functionality
https://github.com/twbs/bootstrap/blob/master/js/tooltip.js

    viewport: {
      selector: 'body',
      padding: 0
    }

bstruthers:
@mgcrea Indeed : )  I've been sitting on it for a while.  The AngularStrap and Bootstrap positioning code were very different.  Some differences between this and #1429 

- Mirrors Bootstrap's implementation.  (This may address other reported issues involving options.container)
- Allows specifying the viewport and padding (like bootstrap)

The big change was matching bs and adding some sanity to the tests : )

gaui:
I recommend #1429 - if he uses `window.innerHeight` and `window.innerWidth` instead of dimensions for the `html` DOM element. It can be modified via CSS, `window` object can not.

mgcrea:
@bstruthers Had some more time to review it, looks good, I like the fact that it mirrors bootstrap's implementation. Would need to bump a minor for it ;-).

bstruthers:
@gaui I'm not sure I understand what your concern is.  Can you elaborate?

RobJacobs:
This change plays into some refactoring I was doing on the helper dimensions module.  I merged this change in with my changes here: https://github.com/RobJacobs/angular-strap/commit/ea88f13c2d94c3dda53c9963f6ef0b486df3339e

What I did:
1. Renamed dimensions to $position to better reflect it's responsibility.
2. Renamed debounce and throttle to $debounce and $throttle so now all helper modules are named consistently.
3. Moved tooltip exotic placement feature to $position module so it can be repurposed in other directives.

There are still some issues I would like an opinion on, see my TODO comments in the helpers.position.js file.  The exotic placement method in the $position module is not taking into account the viewport changes, I would still need to move some pieces of the applyPlacement method in the tooltip down to the $position positionPlacement method as it seems there are some additional position/sizing calcs happening there.  I would like to keep the positionPlacement method as a calc routine only, and have the consuming object be responsible for what to do with the calculated results.

mgcrea:
Landed in c9daa61. Great job @bstruthers!

PR #1477, opened 2015-01-29
display on screen even though body height is greater than screen height

### LINKED ISSUE #1476: Dropdown stay on screen when opened.
Hi im having problems with Dropdown opening on the screen. i created a Stack overflow post with the problem. So you guys have this "auto"  placement abillity which is pretty nice. it tries to keep the dropdown visable. i would like it to work better. if i set the container to body and set placement to auto  it will flip the dropdown to the other side when the dropdown will not fit horizontally and that works fine. however if i'm scrolled into the middle of the page and the dropdown button is very low on the page i would like it to open up but not down so i don't have to scroll down to see it. so i was wondering if 

one solution for this could be to change line 4554 to 4559

	var elPos;
	if (options.container === 'body') {
	  elPos = dimensions.offset(el); // <--- after this line here 4556
	} else {
	  elPos = dimensions.position(el);
	}
to

    (auto) {
        elRect = angular.extend({}, { top: 0, right: window.innerWidth, bottom: window.innerHeight, left: 0 });
    }

you would also maybe want to add a boolean to check if the getPosition call comes from the "auto" placement part of the code

SO post: http://stackoverflow.com/questions/28196724/angular-strap-tooltip-auto-placement?noredirect=1#comment44779159_28196724

Hrafnkellos:
added auto boolean parameter to get possition. if auto is true and container is body function will get screen height and calculate from that if dropdown goes up or down.
 
fixes #1476

### DIFF
--- a/src/tooltip/tooltip.js
+++ b/src/tooltip/tooltip.js
@@ -346,7 +346,7 @@ angular.module('mgcrea.ngStrap.tooltip', ['mgcrea.ngStrap.helpers.dimensions'])
           if (autoPlace) {
             var originalPlacement = placement;
             var container = options.container ? angular.element(document.querySelector(options.container)) : element.parent();
-            var containerPosition = getPosition(container);
+            var containerPosition = getPosition(container, autoPlace);
 
             // Determine if the vertical placement
             if (originalPlacement.indexOf('bottom') >= 0 && elementPosition.bottom + tipHeight > containerPosition.bottom) {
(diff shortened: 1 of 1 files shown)


bstruthers:
@Hrafnkellos I'm curious to know if this will still be needed with PR #1472

Hrafnkellos:
maybe not. if that pull request fixes the problem than thats wonderfull :+1: .

mgcrea:
#1472 has been merged, please chime in if it is still not fixed, closing this for now.

Issue #1484, opened 2015-02-03
$aside fails to load templates that are already included


dinoboff:
It's probably an issue with the modal, but when trying to use a template that might have been already loaded (by ng-include), $aside fails to fetch the template from the cache:

http://plnkr.co/edit/FwZp9RusvPbEuxTA3nLy?p=preview

The template, `foo.html` is first included. When clicking on the button to open the aside, its content will be empty.

If you remove the ng-include and let $aside cache the template, the aside works correctly.

dinoboff:
The problem seems to be there:

https://github.com/mgcrea/angular-strap/blob/master/src/modal/modal.js#L281

Instead of setting and retrieving the cache directly, modal should let $http handle caching.

mgcrea:
You often want to precompile your templates and put them in `$templateCache` using your build tools (cf. https://github.com/mgcrea/gulp-nginclude for instance). `$http` wouldn't properly leverage those.

dinoboff:
@mgcrea Thanks for the fix.

When testing my fix with `$aside`, `$modal` would get `aside.tpl.html` from the cache populated by `gulp-nginclude` and it would get the content of the aside from the template cached by the `ngInclude` directive.

mgcrea:
@dinoboff I'll check that again, we should make sure it works for `1.2` too.
The `fetchPromise` pattern would still be useful as it does prevent parallel `$http` requests when you have multiple directives with a non-cached template on the same page (eg. `ng-repeat` with `bs-tooltip`)

PR #1486, opened 2015-02-03
fix($modal): let $http handle caching

dinoboff:
`$modal` and `$tooltip` are using the same cache than `ngInclude` but are caching templates as object where `ngInclude` is caching it as array (at least in angularjs 1.3).

Fix issue #1484


mgcrea:
Landed in 0357f64ecdf75781452e63b45446ae3b5f47fc45. Thanks!

PR #1487, opened 2015-02-03
enable caretHtml as configurable option for select.js

chrisjbrown:


mgcrea:
Thanks! Landed in f356f0d.

PR #1495, opened 2015-02-10
feat(bsTabs): add disabled option for bsPane

cemeng:
This PR adds a feature where a pane can be disabled by setting disabled attribute on bs-pane. 

Needed this feature on our project where we use bs-tabs for wizard that prevents user from progressing to next tab on certain conditions.

screen shot 2015-02-10 at 2 05 55 pm


mgcrea:
Looks good, thanks for the PR! Landed in df8f02f

PR #1499, opened 2015-02-11
feat(tabs, select) Adding href attribute to select and tab anchor elements.

RobJacobs:
Having an empty href attribute on the anchor element will make the states (hover, focus) appear correctly.  This removes inline cursor style on select option anchor element so it can be overridden.


mgcrea:
@RobJacobs Thanks for the PR but it lacks focus, could you please at least revert the renaming and submit it as another PR? Would ease the review.

RobJacobs:
@mgcrea Sorry about that, only the first commit was related to this issue.  The other 2 were a change proposal for https://github.com/mgcrea/angular-strap/pull/1472.  I rebased my fork and removed the 'rename' commits, but I'd still like to get your opinion on:

1.  Refactoring debounce, throttle, and dimensions modules to append a $ so they are named consistently with the other helper modules.

2.  Move the exotic positioning functionality in tooltip directive to a helper module so it can be leveraged by other directives.

mgcrea:
@RobJacobs,

1. It's pretty subjective, but I like the naming as it is. `debounce` and `throttle` being simple helper functions, and not actual services, they don't "qualify" for a `$` prefix. Moreover, `$` is reserved for the Angular team, so ideally we should move away from them to avoid any future conflicts. I tend to use the `$camelCase` anyway for factories, and `CapitalizedClass` for actual classes.

2. Might be a good idea, but like the backdrop thing with the modal, if you need an appearing element with heavy positioning need, (eg. popover), you can just extend the tooltip. Clearly this extend scheme has its limits and I'm open to suggestions moving towards more modularized providers/directives. My position on this is whether the change would simplify the code, it might, but we would have to refactor everything depending on the tooltip to move from the extend-tooltip-pattern to the require-needed-modules one. Not sure it's worth it so close to Angular2 & the ES6 migration coming. Maybe we should wait to refactor everything for that.

RobJacobs:
@mgcrea Do you still need anything from me on this pull request?

PR #1501, opened 2015-02-11
feat(timepicker): add option to floor minutes display to nearest 5 minut...

### LINKED ISSUE #1444: Ability to set initial minutes value
It would be nice to set the initial minutes value without having to specify a value in the model.

For example, if it is currently 11:42 AM, the default timepicker without a value in the model will popup with:
2015-01-19 11_42_56-dws - 1220

This is due to the code:
```js
var startDate = controller.$dateValue || new Date();
var viewDate = {
    hour: startDate.getHours(), 
    meridian: startDate.getHours() < 12, 
    minute: startDate.getMinutes(), 
    second: startDate.getSeconds(), 
    millisecond: startDate.getMilliseconds()
};
```

Which gets the current date if a date isn't set already.

It would be nice to be able to default the minutes values to 30, 35, 40, 45, 50 by pegging the initial minutes value at 0 rather than at 2.

This doesn't occur if you set a default value in the model, but in some cases this is not desirable.

vmlf01:
...es whole value

Adds a new option 'roundDisplay' that is checked when the initial time value specified by ngModel is not defined. Previously, the timepicker would use the current time in this case. With this new option, we can tell the picker to round the minutes value so we don't get times like 11:38 or 11:43, but instead get 11:35 or 11:40

Closes #1444

### DIFF
--- a/src/timepicker/docs/timepicker.demo.html
+++ b/src/timepicker/docs/timepicker.demo.html
@@ -236,6 +236,14 @@ $scope.sharedDate = {{sharedDate}}; // (formatted: {{sharedDate | date:'short'}}
           </td>
         </tr>
         <tr>
+          <td>roundDisplay</td>
+          <td>boolean</td>
+          <td>false</td>
+          <td>
+            <p>Whether the picker should round the minute values displayed when no initial time is specified. The rounding is made by dividing time in minuteStep intervals and flooring the current time to the nearest.</p>
+          </td>
+        </tr>
+        <tr>
           <td>iconUp</td>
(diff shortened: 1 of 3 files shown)


vmlf01:
Just noticed, maybe this should use the minuteStep option value instead of being fixed to 5 minutes.
Will update PR tomorrow.

vmlf01:
I've corrected the calculations to take into account the value of minuteStep, please review it whenever you can.

mgcrea:
Thanks! Landed in bd2ace3.

Issue #1410, opened 2015-01-07
Modal not getting along with class="col-sm-2"


AdamGorkoz:
It seems like any element inside an element with class="col-sm-2" is not getting darkened when modal shows. 
![image](https://cloud.githubusercontent.com/assets/10440278/5652643/fa6d91b0-96ba-11e4-834d-df859635b60d.png)
P.S
After checking with other col-sim sizes i see the problem still occurs.

Regards,
Turka50.

vmlf01:
I think this was caused by the changes in Bootstrap 3.3.x releases to the modal behavior so it needs some fixing to work with all versions.

turka50:
Any suggestions how this fix issue locally ?

vmlf01:
Can you go back to Bootstrap 3.2 until the issue is fixed?

vmlf01:
@turka50 please check again if problem is fixed with next release of angular-strap. (includes PR #1508)

PR #1508, opened 2015-02-15
fix(backdrop): force modal backdrop position and z-index (fixes #1263)

### LINKED ISSUE #1263: Mimic the new behavior of modal backdrop (Bootstrap v3.3.0 update)
Following https://github.com/twbs/bootstrap/issues/15066#issuecomment-62403892 backdrop of modals are not visible due to the change and removal of z-index in original Bootstrap library.

dankahle:
Force the modal positioning and z-index to overcome bootstraps 3.3.0 change of location.

Fixes #1263

### DIFF
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -86,6 +86,7 @@ angular.module('mgcrea.ngStrap.modal', ['mgcrea.ngStrap.helpers.dimensions'])
         // Fetch, compile then initialize modal
         var modalLinker, modalElement;
         var backdropElement = angular.element('<div class="' + options.prefixClass + '-backdrop"/>');
+        backdropElement.css({position:'fixed', top:'0px', left:'0px', bottom:'0px', right:'0px', 'z-index': 1039});
         $modal.$promise.then(function(template) {
           if(angular.isObject(template)) template = template.data;
           if(options.html) template = template.replace(htmlReplaceRegExp, 'ng-bind-html="');


dankahle:
This came to mind to me this morning: it's just a backdrop and just because they put theirs inside their .modal, doesn't mean we have to. The whole problem is the absolute positioning and z-index. Maybe naive of me, but if nothing else, a temporary fix to something that's been broken since november. Anyway, hadn't seen anyone suggesting this. It's an easy fix, works with both 3.2.0 and 3.3.2, the new .modal z-index was moved to 1040, had been 1040/1050 for backdrop/.modal so just skirted under the .modal at 1039.

RobJacobs:
I would like to see this as a separate directive or service, would be a useful component to have available to other directives.

mgcrea:
@dankahle I think this is probably the way to go, have the backdrop bundled in the template brings quite a few issues. At first I was really eager to match bootstrap behavior. But after having tried to fix it a couple of times without finding a good enough solution, it's time to give up for now.

@RobJacobs not sure about the separate directive, could be interesting if it simplifies code but I'm not sure this is really feasible. Modal is the base class for backdrop-enabled elements, so any directive needing the behavior (eg. aside) can just extend it.

RobJacobs:
@mgcrea I've been doing some work on implementing a filter for the select directive and I'm struggling the tooltip trigger events. Would be nice if the tooltip had a dismiss trigger like the modal backdrop click. I would also like something like that for the popover. I was thinking if we could abstract that functionality from the modal it could be shared with the tooltip.

vmlf01:
This might fix #1410 also.

PR #1509, opened 2015-02-16
fix(modal directives): make falsey boolean options work

dankahle:
This continues/extends the work @vmlf01  was doing on #1462. I'm going to go through all the directives on this issue, just thought I'd PR these three as they're related (all modal) and get a confirmation that the solution looks nice. I found a nice clean way of doing it: set the values, then overwrite the falsy booleans, should there be any. I was going to push for attribute existence being "true", but ran into the "container" option (was missed by last PR) where it expects all truthy values to be usable (as a dom query value), true breaks that code, so just ended up keeping the existing truthy values, but overwriting if falsy (false, 0, empty). I think it looks nice. Let me know. I'll be working on the rest of them tomorrow. 

While I was on the alert, couldn't get the keyboard option to work (hit escape to make it hide). Seems it never has worked? Googled it and found setting tabindex was the key. Problem is: made the stupid chrome outline show up as it has focus. Tried style="outline:none" to no avail, so went with outline-width:0px. Worked. Kinda hacky, but that outline looks horrible.

### DIFF
--- a/src/alert/alert.js
+++ b/src/alert/alert.js
@@ -80,7 +80,7 @@ angular.module('mgcrea.ngStrap.alert', ['mgcrea.ngStrap.modal'])
 
         // use string regex match boolean attr falsy values, leave truthy values be
         var falseValueRegExp = /^(false|0|)$/i;
-        angular.forEach(['keyboard', 'html', 'container'], function(key) {
+        angular.forEach(['keyboard', 'html', 'container', 'dismissable'], function(key) {
           if(angular.isDefined(attr[key]) && falseValueRegExp.test(attr[key]))
             options[key] = false;
         });
--- a/src/alert/alert.js
+++ b/src/alert/alert.js
@@ -78,6 +78,13 @@ angular.module('mgcrea.ngStrap.alert', ['mgcrea.ngStrap.modal'])
           if(angular.isDefined(attr[key])) options[key] = attr[key];
         });
 
+        // use string regex match boolean attr falsy values, leave truthy values be
+        var falseValueRegExp = /^(false|0|)$/i;
+        angular.forEach(['keyboard', 'html', 'container'], function(key) {
+          if(angular.isDefined(attr[key]) && falseValueRegExp.test(attr[key]))
+            options[key] = false;
+        });
+
         // Support scope as data-attrs
         angular.forEach(['title', 'content', 'type'], function(key) {
           attr[key] && attr.$observe(key, function(newValue, oldValue) {
--- a/src/alert/docs/alert.demo.html
+++ b/src/alert/docs/alert.demo.html
@@ -105,12 +105,16 @@
           <td>'info'</td>
           <td>default content value if <code>data-type</code>attribute isn't present</td>
         </tr>
+<!--
+Removing keyboard from the docs. Our template doesn't have tabindex for focus stealing issues, keyboard won't work
+so remove the option then. Just looks bad to have an option that doesn't work
         <tr>
           <td>keyboard</td>
           <td>boolean</td>
           <td>true</td>
           <td>Closes the alert when escape key is pressed</td>
         </tr>
+-->
         <tr>
           <td>show</td>
           <td>boolean</td>
--- a/src/aside/aside.js
+++ b/src/aside/aside.js
@@ -54,6 +54,13 @@ angular.module('mgcrea.ngStrap.aside', ['mgcrea.ngStrap.modal'])
           if(angular.isDefined(attr[key])) options[key] = attr[key];
         });
 
+        // use string regex match boolean attr falsy values, leave truthy values be
+        var falseValueRegExp = /^(false|0|)$/i;
+        angular.forEach(['backdrop', 'keyboard', 'html', 'container'], function(key) {
+          if(angular.isDefined(attr[key]) && falseValueRegExp.test(attr[key]))
+            options[key] = false;
+        });
+
         // Support scope as data-attrs
         angular.forEach(['title', 'content'], function(key) {
           attr[key] && attr.$observe(key, function(newValue, oldValue) {
--- a/src/aside/test/aside.spec.js
+++ b/src/aside/test/aside.spec.js
@@ -3,6 +3,7 @@
 describe('aside', function () {
 
   var $compile, $templateCache, scope, sandboxEl;
+  var bodyEl = $('body');
 
   beforeEach(module('ngSanitize'));
   beforeEach(module('mgcrea.ngStrap.aside'));
@@ -43,6 +44,9 @@ describe('aside', function () {
     'options-template': {
       scope: {aside: {title: 'Title', content: 'Hello aside!', counter: 0}, items: ['foo', 'bar', 'baz']},
       element: '<a data-template="custom" bs-aside="aside">click me</a>'
+    },
+    'options-backdrop': {
+      element: '<a data-placement="center" bs-aside="aside" data-backdrop="0">click me</a>'
     }
   };
 
@@ -171,6 +175,24 @@ describe('aside', function () {
       });
 
     });
+
+    describe('backdrop', function() {
+      it('should show backdrop by default', function() {
+        var elm = compileDirective('default');
+        expect(bodyEl.find('.aside-backdrop').length).toBe(0);
+        angular.element(elm[0]).triggerHandler('click');
+        expect(bodyEl.find('.aside-backdrop').length).toBe(1);
+      });
+
+      it('should not show backdrop if option value if falsy', function() {
+        var elm = compileDirective('options-backdrop');
+        expect(bodyEl.find('.aside-backdrop').length).toBe(0);
+        angular.element(elm[0]).triggerHandler('click');
+        expect(bodyEl.find('.aside-backdrop').length).toBe(0);
+      });
+
+    });
+
 
   });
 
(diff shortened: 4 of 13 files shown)


dankahle:
Also, when going through the alert I noticed it doesn't have a "html" setting and has ng-bind-html coded right into the template, this doesn't seem right as not even the tooltip does that. I left the html setting in there in case that would be pulled out in the future, it affects nothing as it can only add what's already there.

mgcrea:
@dankahle Harmonizing behavior across directive sounds good. Indeed the alert `ng-bind-html` looks fishy, it's built upon the modal though, but the modal does handle the html option (https://github.com/mgcrea/angular-strap/blob/master/src/modal/modal.js#L92).

dankahle:
I'm seeing a lot of fishy as I go through these. I'm amassing a list of questions for you to make a decisions on. I'll throw them in these comments in an hour or so.

dankahle:
Here's my questions:

* I’ve been using a case insensitive regex (false/0/nothing) as I didn’t like it calling “False” true, makes sense
* how often do you refresh gh-pages? I see stuff in the docs, but they’re not on the site. All I see in gh-pages is one commit from 2/14/15, but can’t find any changes in that index.html file even though there’s been changes to the docs index.html? I’m confused.

 Alert
* no html in docs and ng-bind-html in code. Is this what you wanted? If so, why this widget and not the others? From what I can see all widgets have html settings (though not sure the actually work), this seem the only one that doesn’t that I know of. It uses the modal, so all the code is there, just not in the docs and already in the template. What to do?
* Had to add tabindex to get the keyboard working. I’m still at a loss as to why the chrome focus outline was an issue with this (I had to set outline-width:0px to hide it) and not any of these other template that all have tabindex as well. Don’t see any special outline settigns in the bootstrap css for them. 

tooltip
* keyboard not in docs, but in attr accept code (so can be set). No tabindex in template and set to false by default.  Do we want keyboard or not here?

popover
* keyboard not in docs, but in attr accept code (so can be set). No tabindex in template and set to "true" by default. The docs eg with “your own template” has tabindex and works with keyboard.  Do we want keyboard here or not?

datepicker
* keyboard?? This woudl be a great feature, but it appears it already works in docs, but without setting keyboard. Need to check into it. Maybe it’s autoclose and that closes on escape? Anyway with autoclose=false, might still be nice, as when they’re done with it, they can just hit escape.

timepicker
* roundDisplay was added to the code and docs, but it’s not in gh-pages. When will that happen?

buttons
this code is way different. Gonna have to go through it later

tabs
this code is different, will have to go through later

mgcrea:
> how often do you refresh gh-pages? I see stuff in the docs, but they’re not on the site. All I see in gh-pages is one commit from 2/14/15, but can’t find any changes in that index.html file even though there’s been changes to the docs index.html? I’m confused.

`gh-pages` are built using `gulp` with the `gulp pages`, usually, I would only update `gh-pages` when a new release comes out. I might sometimes push intermediate builds but it's not really the norm. Currently the build is broken, `gulp-usemin` breaks, I've been wanting to move the build workflow to a more robust/modern solution for a long time (mainly leveraging the ongoing work on [ng-factory](https://github.com/ng-tools/ng-factory)), migrating the whole gulp workflow is pretty complex, but having the `gulp pages` breaking made me decide to switch, I'm currently half-way (dist is OK but pages are not).

> I’ve been using a case insensitive regex (false/0/nothing) as I didn’t like it calling “False” true, makes sense?

Yes, it should be insensitive

> Alert, no html in docs and ng-bind-html in code. Is this what you wanted? If so, why this widget and not the others? From what I can see all widgets have html settings (though not sure the actually work), this seem the only one that doesn’t that I know of. It uses the modal, so all the code is there, just not in the docs and already in the template. What to do?

Not really, I guess I might have skipped it during a refactoring of the html option. Never had any issue with it but it should be patched to be consistant. 

> Alert, Had to add tabindex to get the keyboard working. I’m still at a loss as to why the chrome focus outline was an issue with this (I had to set outline-width:0px to hide it) and not any of these other template that all have tabindex as well. Don’t see any special outline settigns in the bootstrap css for them.

Had to remove the tabindex to prevent focus steal (see https://github.com/mgcrea/angular-strap/commit/ad9d0855fcd2f59ad1904db7ca7f8d031992b198). Modal-like alerts are exclusive to AngularStrap but we could add CSS code to [Bootstrap Additions](https://github.com/mgcrea/bootstrap-additions). 

> Tooltip, keyboard not in docs, but in attr accept code (so can be set). No tabindex in template and set to false by default. Do we want keyboard or not here?

Tooltip is a base class used by other components, so it may have code not used by default for it.

> Popover, keyboard not in docs, but in attr accept code (so can be set). No tabindex in template and set to "true" by default. The docs eg with “your own template” has tabindex and works with keyboard. Do we want keyboard here or not?

I think it might depend on your use case, a popover can act as a larger tooltip, but it can also be a form... Maybe we should link the tabindex attribute to the keyboard option being truthy? 

> datepicker, keyboard?? This woudl be a great feature, but it appears it already works in docs, but without setting keyboard. Need to check into it. Maybe it’s autoclose and that closes on escape? Anyway with autoclose=false, might still be nice, as when they’re done with it, they can just hit escape.
timepicker

I think it's autoclose with escape just changing the focus.

> roundDisplay was added to the code and docs, but it’s not in gh-pages. When will that happen?

See 1.

The code is pretty old and you might find other oddities... Feel free to dig them and submit PRs!

dankahle:
##### alert: fix html option to work again and put ng-bind back in the template:
problem is: "html" option has been removed from the docs, so they won't know how to turn it on. The docs don't get updated so who knows when they'll see the option. Maybe create an issue documenting it saying it's waiting for the docs update. This could break people's code, then no doc's option to show them how to turn it back on, not cool.

##### alert tabindex add: 
I'll pull the tab index back out then, then the keyboard option is meaningless, so I'll pull that out of the docs as well. Maybe just comment it out with the reason why in the comments.

##### tooltip/popup/datepicker keyboard issues:
you've got me pretty gunshy with the focus-stealing thing going on. Think I'll steer clear for now, though I do like the idea of adding tabindex if keyboard is truthy. Problem is: they won't be able to connect the dots: keyboard=true >> focus steal. They'll just think it's broken. All three don't have keyboard option in the docs. I'm thinking this is by design. I'm gonna leave it be then.

dankahle:
Gonna wrap this with tests, but could take awhile as I'll need to come up to speed on the tests for all the widgets, and all these options, a daunting task. Not sure if there's a shortcut. If I can manage to spy on the service and just make sure it's called with the right options, that could save a lot of tests, but then we wouldn't be testing the service actually following through with the command. If I test each case (truthy, "", False, 0) for each option on each widget, that will junk up the tests. It's more like I need one truthy and one falsy integration test for each option (still a lot) and then spy on the service and hit it with all the permutations to see if the service is called with them. Gonna take me awhile on this, and these options have  been broke since the beginning I assume.

ansonkao:
Nice find + fix!

The only criticism I have around this solution is that it'll be easy for future maintainers to miss this step.
I banged my head on the table for a bit, wondering why my unit test wasn't working on a new option I added in #1532. I guess it's one of the inherent pitfalls of JS not having typesafety. 

What if we commented each option in the defaults object like so:
...
someOption: "default",
someOtherOption: false, // Falsy option - see https://github.com/mgcrea/angular-strap/pull/1509
SomeThirdOption: false, // Falsy option - grep this page for #falsyFix
...

Thoughts?

NTS: add this fix to #1532 after both PR's are merged in.

dankahle:
Tooltip.js
How does this watch('bstooltip") code not throw an sce exception? The content isn't made safe in the code, yet it works in the docs. It fails in the test over an sce error, which makes sense as it's not being marked sce.safeAsHtml and my data-html setting to true has changed the template to ng-bind-html. 

.directive('bsTooltip'):

        attr.bsTooltip && scope.$watch(attr.bsTooltip, function(newValue, oldValue) {
          if(angular.isObject(newValue)) {
           >>>>>>  angular.extend(scope, newValue);
          } else {
            >>>>>>  scope.title = newValue;
          }
          angular.isDefined(oldValue) && $$rAF(function() {
            tooltip && tooltip.$applyPlacement();
          });
        }, true);

dankahle:
tooltip autoclose option:
is there a reason for there NOT being a tooltip autoclose option (in docs and directive) but popover does, even though it's the same tooltip code? Just curious.

ansonkao:
@dankahle why did you close this? I thought this was a good idea.

dankahle:
I wasn't aware that I closed anything. What got closed? your comment? I see the red closed circle with my pic next to it :). Guilty as charged I guess. Didn't know you could close comments. I'm trying to get out of this one after 4 or 5 days work on it. Not really getting what you're wanting (above). That the default options be labeled? My concern is the boolean "false" and "0" values that were being called truthy by accident. This stuff is a bit confusing. The boolean values are easy enough and my code appends the attribute settings with a special boolean handling, which is hard to miss. The docs do a nice enough job of stating what all the values "can" be, and the defaults are all stated in an object in the directive/factory. 

I should be wrapping this up today.

ansonkao:
@dankahle awesome!

I was referring to the fact that this solution requires each boolean option to be listed in that boolean attributes array, and future maintainers may not realize this when they add new options. So they may add a new boolean option to the defaults without adding it to the boolean array, and the problem will leak back in.

dankahle:
I guess my force push closed this pull request? Seems you have to force push when you rebase on the latest master. I'm still learning this git thing. I'm done with this pull request finally. Was trying to get all my changes into one commit, but ran into merge conflicts last time I rebased and it created a commit for that. Not sure what to do about that.

dankahle:
This PR is still closed after the last "reopen and comment". What should I do now? Create a new PR? I guess so.

PR #1519, opened 2015-02-18
fix(modal backdrop): make z-index not conflict with aside

dankahle:
The bootstrap-additions code sets the aside z-index to bootstraps: @z-index-modal -1 which will be 1039 the next time you update using bootstrap 3.3.2. 1039 is the value I chose for the modal backdrop in 81f8654a9e92341eaf4d495ec132a1e6cf7743ca , so I'm moving the backdrop to 1038 to stay under aside.


mgcrea:
Thanks, landed in eb0635e.

PR #1524, opened 2015-02-22
fix(select): don't process navigation keys with multiple select

### LINKED ISSUE #1521: Multiple select stops working when arrow keys used for navigation.
This issue occurs when arrow keys are used to navigate in multiple select, its replicable on http://mgcrea.github.io/angular-strap/#/selects#selects.

This error is being thrown on console :
Error: scope.$activeIndex.indexOf is not a function SelectFactory/$select.$isActive@http://mgcrea.github.io/angular-strap/dist/angular-strap.js:2993:20 SelectFactory/scope.$isActive@http://mgcrea.github.io/angular-strap/dist/angular-strap.js:2912:18 hb.prototype.functionCall/<@http://code.angularjs.org/1.3.8/angular.min.js:198:422 h/<@http://code.angularjs.org/1.3.8/angular.min.js:108:158 Pe/this.$get</l.prototype.$digest@http://code.angularjs.org/1.3.8/angular.min.js:123:308 safeDigest@http://mgcrea.github.io/angular-strap/dist/angular-strap.js:4657:66 TooltipFactory/$tooltip.hide@http://mgcrea.github.io/angular-strap/dist/angular-strap.js:4351:11 SelectFactory/$select.hide@http://mgcrea.github.io/angular-strap/dist/angular-strap.js:3061:11 TooltipFactory/$tooltip.leave@http://mgcrea.github.io/angular-strap/dist/angular-strap.js:4321:1 lf/c@http://code.angularjs.org/1.3.8/angular.min.js:32:358

The problem here is that in case of multiple select the scope.$activeIndex is an array, but due to this arrow keys navigation code at $select.$onKeyDown function: 

          // Navigate with keyboard
          if(evt.keyCode === 38 && scope.$activeIndex > 0) scope.$activeIndex--;
          else if(evt.keyCode === 40 && scope.$activeIndex < scope.$matches.length - 1) scope.$activeIndex++;
          else if(angular.isUndefined(scope.$activeIndex)) scope.$activeIndex = 0;

The array scope.$activeIndex gets converted to a number and so the above error comes scope.$activeIndex.indexOf is not a function. So here either the navigation should not be allowed and the condition should be

```
if(!options.multiple) {
      // Navigate with keyboard
      if(evt.keyCode === 38 && scope.$activeIndex > 0) scope.$activeIndex--;
         else if(evt.keyCode === 40 && scope.$activeIndex < scope.$matches.length - 1)     scope.$activeIndex++;
         else if(angular.isUndefined(scope.$activeIndex)) scope.$activeIndex = 0;
}

```

Or if navigation is to be allowed in multiple select, then there needs to be a separate index and would require some more code changes and maybe a new color to distinguish between selected items and the current index.

vmlf01:
Keyboard navigation/selection is not implemented for multiple selects, so we need to check if select type is not multiple before handling cursor keypresses.
This should really be handled by implementing the keyboard navigation, but for now, just a quick fix.

Fixes #1521

### DIFF
--- a/src/select/select.js
+++ b/src/select/select.js
@@ -189,11 +189,13 @@ angular.module('mgcrea.ngStrap.select', ['mgcrea.ngStrap.tooltip', 'mgcrea.ngStr
             return $select.select(scope.$activeIndex);
           }
 
-          // Navigate with keyboard
-          if(evt.keyCode === 38 && scope.$activeIndex > 0) scope.$activeIndex--;
-          else if(evt.keyCode === 40 && scope.$activeIndex < scope.$matches.length - 1) scope.$activeIndex++;
-          else if(angular.isUndefined(scope.$activeIndex)) scope.$activeIndex = 0;
-          scope.$digest();
+          if (!options.multiple) {
+            // Navigate with keyboard
+            if(evt.keyCode === 38 && scope.$activeIndex > 0) scope.$activeIndex--;
+            else if(evt.keyCode === 40 && scope.$activeIndex < scope.$matches.length - 1) scope.$activeIndex++;
+            else if(angular.isUndefined(scope.$activeIndex)) scope.$activeIndex = 0;
+            scope.$digest();
+          }
         };
 
         // Overrides


mgcrea:
Thanks! Landed in 516b67b.

Issue #1480, opened 2015-01-29
data-placement="top-left" on Typehead show incorrectly


nathanredblur:
Bad location on use typehead with top placement

here the example
http://plnkr.co/edit/GEJN2PuB721t3NdspS8z?p=preview

also, how works data-placement="auto" ?
because I think that if the dropdown is out of the viewport appear in top location.

bstruthers:
This issue is caused by how the $tooltip service initially places the results list. When the placement routine executes, the `$scope` is still in the middle of a `digest` and the list hasn't been populated with its content.  Since the height is essentially 0, the service places the list at the very top of the input.  Once the digest phase completes, the content is rendered but the list doesn't reposition itself.  

Wrapping the placement in a $$postDigest call does get the list to position correctly:
![placement](https://cloud.githubusercontent.com/assets/124591/6307519/36457720-b902-11e4-8cf8-9f9d8cbd2985.png)

but as you refine the search results, the placement doesn't adjust itself:
![placement2](https://cloud.githubusercontent.com/assets/124591/6307543/6115506a-b902-11e4-99e7-623e00df7b9e.png)

The typeahead would need to be updated to reposition itself whenever the content changes.

As for the question of auto, it's not tied to the viewport.  The changes to support viewport positioning are in 2.2.0.  Auto attempts to place the element, and if that placement falls outside the bounds of its container (default of parent), it uses the opposite placement.  

For example, if I set the placement to `auto top` on the state typeahead, it attempts to place the list above above the input.  Since placement causes it to be outside the `bs-example` div, it's placement is swapped to bottom.

PR #1525, opened 2015-02-23
fix(typeahead): reposition results on update

bstruthers:
When the placement is not one of the bottom placements, the typeahead is positioning the results incorrectly.

Calling the $applyPlacement function inside the typeahead's update function, will cause the results to position correctly as the suggestions are updated.  This fixes issue https://github.com/mgcrea/angular-strap/issues/1480


mgcrea:
Thanks for the well crafted PR, landed in 109e7e0.

PR #1526, opened 2015-02-23
Include role and aria-controls for better tab accessibility

kkirsche:
Include `role="presentation"` and `aria-controls` per http://getbootstrap.com/javascript/#tabs


mgcrea:
Thanks! Landed in 3d7ed73.

PR #1527, opened 2015-02-23
Use aria attributes in dropdown demo

kkirsche:
Use aria attributes in dropdown demo for better accessibility


mgcrea:
Thanks! Landed in efb2f78.

PR #1528, opened 2015-02-23
Improve modal accessibility with ARIA

kkirsche:
Improve modal accessibility with ARIA per the examples on http://getbootstrap.com/javascript/#modals-examples


mgcrea:
Landed in b2dd8a9.

PR #1529, opened 2015-02-23
Use ARIA in collapse demo

kkirsche:
Use ARIA in collapse demo per http://getbootstrap.com/javascript/#collapse-example-accordion


mgcrea:
Thanks landed in 1460bef.

PR #1530, opened 2015-02-23
Use nav instead of div in navbar example

kkirsche:
Use `nav` instead of `div` in example per http://getbootstrap.com/components/#navbar


mgcrea:
Thanks, landed in 5205f09. Beware of commit message guidelines for future PRs.

vmlf01:
For IE 8 or bellow this will need a shim as ```nav``` is HTML5

kkirsche:
If you feel a shim is necessary based on the demographics for Angular-Strap, I would recommend a simple:

```html
<!--[if lt IE 9]>
<script src="https://raw.githubusercontent.com/aFarkas/html5shiv/master/dist/html5shiv.js"></script>
<![endif]-->
```

With that said, considering Angular-Strap is for Bootstrap 3 and it would require Respond.js to work correctly with IE8 anyway, I personally feel a shim is not something that should be included.

@mgcrea Sorry about the commit guidelines. Missed that.

Issue #1522, opened 2015-02-20
Tooltip prevents ng-click on buttons on mobile devices


j-ulrich:
Using the `bs-tooltip` directive on a button prevents that `ng-click` is triggered on mobile devices.
Here is a Plunker demonstrating the issue:
http://plnkr.co/edit/nzcCe6PJno7pNRgnwTxk?p=preview

Works fine when clicking the button with the mouse but using a touch, only the tooltip is shown and the button action cannot be triggered.
The issue can be fixed by setting the trigger to `'hover'` (e.g. `data-trigger="hover"`).

I saw that there is a special case for this situation in [`bindTriggerEvents()`](https://github.com/mgcrea/angular-strap/blob/master/src/tooltip/tooltip.js#L426) but since `$tooltip.$onFocusElementMouseDown()` calls [`evt.preventDefault();`](https://github.com/mgcrea/angular-strap/blob/master/src/tooltip/tooltip.js#L411), this makes buttons unusable on mobile devices.

bstruthers:
Seems like the easiest fix is to change from `touchstart` to `touchmove`.  See http://plnkr.co/edit/92hSukMZQuVaxAgPgMsO?p=preview.  I'm basing this off of: https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Touch_events#Handling_clicks

I've tried the fix out with Chrome and Firefox on Android.  If it looks like a viable solution, I'll pull together the PR.

PR #1531, opened 2015-02-23
fix(tooltip): use touchend instead of touchstart

bstruthers:
When implementing the focus trigger on mobile devices, use touchend so the touchstart event can still propagate and trigger any click event listeners.  Fixes https://github.com/mgcrea/angular-strap/issues/1522

### DIFF
--- a/src/typeahead/test/typeahead.spec.js
+++ b/src/typeahead/test/typeahead.spec.js
@@ -278,6 +278,11 @@ describe('typeahead', function () {
       var elm = compileDirective('markup-objectValue', {}, function(scope) { scope.selectedIcon = 'Ge' });
       expect(elm.val()).toBe('Ge'); // display value will be undefined, use model value for display
     });
+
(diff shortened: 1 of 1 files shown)


mgcrea:
Have you tested it on iOS/Android? Does it adds a slight delay?

bstruthers:
I tested on Android 5.0 with Chrome and Firefox, I don't have access to an
iOS device.

The delay would only be caused by the user maintaining the touch on the
screen. There is no built in 300ms delay.

I believe the fastclick library uses touchend to trigger a click (with
considerable more code and options)- "FastClick uses document.createEvent
to fire a synthetic click event as soon as touchend is fired by the
browser."

bstruthers:
@mgcrea I'm back playing around with this... what I can't seem to figure out is why the code is disabling the event propagation inside $onFocusElementMouseDown.  Thoughts?  The better fix would be to just remove those two lines.

bstruthers:
@mgcrea Hah, nevermind... I need more coffee this morning!!

PR #1532, opened 2015-02-24
feat(typeahead) New option disableAutoSelect

### LINKED ISSUE #1300: typeahead: first value gets selected when suggestion list is still hidden
I think I already saw this issue somewhere in here, but can't locate it now.

If you go into the docs typeahead for State and just press Enter after focusing the input, "Alabama" will be selected because it is the first value in the list, but we haven't typed anything yet and minLength is not 0, so this doesn't seem right.

ansonkao:
Allow override of the default behaviour of auto-selecting the first match. Non-breaking to prior behaviour.

Fixes #1300

### DIFF
--- a/src/typeahead/docs/typeahead.demo.html
+++ b/src/typeahead/docs/typeahead.demo.html
@@ -32,7 +32,7 @@ $scope.selectedAddress = "{{selectedAddress}}";
       </div>
       <div class="form-group">
         <label>Icon</label>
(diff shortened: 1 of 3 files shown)


vmlf01:
While this might fix #1300 as a side effect, it doesn't seem right. Typeahead should not select the first element if you haven't typed anything yet. I think this should be by default, not an option you have to specify.

ansonkao:
100% agree, just assumed we wanted non-breaking changes, with so many existing users being happy with the behaviour thus far. 

Very easy to flip the script (haha!!) if we have consensus!

mgcrea:
Let's break it.

mgcrea:
Thanks! Landed in fcf0bad.

PR #1533, opened 2015-02-24
refactor(select, typeahead): use bs-options

### LINKED ISSUE #1473: Incorrect Use of NgOptions throughout Angular-Strap directives
Hey,

I was taking a look at the implementation of Angular-Strap and into how it uses NgOptions and found that it is incorrect. While not a major issue in 1.2.x/1.3.x (as it doesn't throw a fatal error when this occurs) it does throw fatal errors in 1.4.x which stops their execution as now it is more strict about requirements and ensuring they are there.

Specifically, NgOptions directive requires the `select` controller which is found on `<select>` elements.

This can be seen in 1.3.x here:
https://github.com/angular/angular.js/blob/v1.3.x/src/ng/directive/select.js#L202

and in 1.4.x here:
https://github.com/angular/angular.js/blob/master/src/ng/directive/ngOptions.js#L325

This incorrect use will throw the following error on the 1.4.x branch:
` Error: [$compile:ctreq] Controller 'select', required by directive 'ngOptions', can't be found!` because it's being used for non-select elements.

The documentation they point to for this is:
https://docs.angularjs.org/error/$compile/ctreq?p0=select&p1=ngOptions

bstruthers:
With Angular 1.4.x, the NgOptions directive requires the select controller.

Update the select and typeahead directives to use bs-options.  Fixes #1473.

### DIFF
--- a/src/datepicker/test/datepicker.spec.js
+++ b/src/datepicker/test/datepicker.spec.js
@@ -55,16 +55,16 @@ describe('datepicker', function() {
       element: '<input type="text" ng-model="selectedDate" ng-required="true" bs-datepicker>'
     },
     'options-animation': {
-      element: '<div class="btn" data-animation="am-flip-x" ng-model="datepickeredIcon" ng-options="icon.value as icon.label for icon in icons" bs-datepicker></div>'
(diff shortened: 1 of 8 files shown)


mgcrea:
Landed in 6bb572e. Thanks!

PR #1534, opened 2015-02-24
fix(datepicker): remove table-layout css from template

### LINKED ISSUE #1506: date picker year/month mode width issue in IE
When using the datepicker in IE11, switching from mode 0 (month/day view) to mode 1 (year/month view), the width of the content is compressed. If you close the datepicker with mode 1 active, then reopen it, the content sizes to the full width of the container. This only happens in IE. Here is a plunkr to demonstrate:
http://plnkr.co/edit/0TzI0E?p=preview
2-13-2015 10-03-17 am
2-13-2015 10-10-03 am

bstruthers:
table-layout="fixed" is causing rendering issues in IE11 when changing from month/day view to year/month view.

The styling is unnecessary because the table's width is set to 100%.  Fixes #1506


bstruthers:
Issue was marked as resolved.  I'm going to close this PR

PR #1541, opened 2015-02-27
fix(select): Does not gain selection until up/down key pressed

### LINKED ISSUE #1443: Single-select select control highlights first item when nothing has been selected
image

Despite the placeholder being visible, the first item in the dropdown appears selected. For multiselect, the behavior is different, and nothing is highlighted initially in the dropdown. The latter behavior is preferable.

ansonkao:
The prior behavior was that the first option was automatically matched upon focus. This patch removes that auto matching, and instead we allow the user to initiate the first match by either pressing the up arrow key or the down arrow keys (in addition to clicking).

Fixes #1443


mgcrea:
Thanks! Landed in 6f2d786.

PR #1545, opened 2015-02-28
fix(alert): prevent title option value to be inherited from parent scope

### LINKED ISSUE #910: $alert title inheriting from $rootScope
In my application, I maintain a variable 'title' in the $rootScope, so that I dynamically modify the page title. If I issue an $alert without specifying the title, it takes the title from the $rootScope, which I don't want. Anyway to prevent this?

The forced way I used to work is define 
title: '(space)'

vmlf01:
Fixes #910


vmlf01:
I see that this was already done in tooltip. Should this be added to modal/popover also?

Issue #1493, opened 2015-02-07
bsAffix Issue After changing routes with ngRoute


jsteinshouer:
On initial page load my affix element seems to pin to the correct position but after changing routes using ngRoute I have to scroll half way down the page before my element appears. Any ideas what I may be doing wrong. Here is the markup I am using.

```
 <div class="bs-sidebar" data-offset-top="-60" bs-affix>
```

ELJK:
I am having the same issue. Help would be much appreciated.

fvisser:
We encountered the same problem. The underlying problem is that the bsAffix directive is triggered before the DOM is rendered after a route change. Because of this the offsets are not calculated correctly. Just wrapping the postlink code in a timeout solves the problem. 

We will create a pull request with the fix

ntvsx193:
Hm... I'm think, used `$timeout` bad idea.

CyborgMaster:
`$timeout` is the only way I know of to delay execution until after the dom has rendered.  In this case it is vital, because there is no way to know the initial offset of the element until after the page has rendered.  I'm have to use this branch to get things to work with ngRoute.  I vote to have this merged.

CyborgMaster:
I have fully tested this patch and found it to work great.  I would love to get it merged because this is a change to only the source file.  I can't use the fork directly because the compilation into `dist` hasn't been run on the fork (as it shouldn't be) and so the main js file hasn't been modified to reflect the patch.  For now, I've made my own fork and manually applied this little patch to the dist file directly.

PR #1555, opened 2015-03-05
Update affix.js

fvisser:
This is a fix for issue #1493. Wrapped the postLink code in a timeout to make sure that the directive is executed after the DOM is rendered to make sure that the offsets are calculated correctly

### DIFF
--- a/src/affix/affix.js
+++ b/src/affix/affix.js
@@ -220,30 +220,30 @@ angular.module('mgcrea.ngStrap.affix', ['mgcrea.ngStrap.helpers.dimensions', 'mg
 
   })
 
-  .directive('bsAffix', function($affix, $window) {
+  .directive('bsAffix', function($affix, $window, $timeout) {
 
     return {
       restrict: 'EAC',
       require: '^?bsAffixTarget',
       link: function postLink(scope, element, attr, affixTarget) {
+        $timeout(function() {
+          var options = {scope: scope, target: affixTarget ? affixTarget.$element : angular.element($window)};
+          angular.forEach(['offsetTop', 'offsetBottom', 'offsetParent', 'offsetUnpin', 'inlineStyles'], function(key) {
+            if(angular.isDefined(attr[key])) {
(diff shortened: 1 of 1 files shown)


CyborgMaster:
I have fully tested this patch and found it to work great. I would love to get it merged because this is a change to only the source file. I can't use the fork directly because the compilation into dist hasn't been run on the fork (as it shouldn't be) and so the main js file hasn't been modified to reflect the patch. For now, I've made my own fork and manually applied this little patch to the dist file directly.

CyborgMaster:
@fvisser, it looks like @mgcrea is asking for the merge conflicts to be resolved and some unit tests to be written.  It that something you can take care of?

DaleWebb:
@CyborgMaster I guess if you can fix the tests and resolve the conflicts, this PR could be ready to go ahead

CyborgMaster:
I tried to get the tests running but I can't even get gulp to run.  Trying to install the dependencies with `npm install` took ~15 minutes, but even when it finished, running `gulp` gave `Error: Cannot find module 'undertaker-lib-tasks'`.  Any ideas?

DaleWebb:
That's strange. Have you tried checking out master and installing and testing?

CyborgMaster:
I was actually testing off of master.  I did finally get gulp to work, but the npm install took over 30 min.  This is on a 2 year old MacBookProRetina, 16Gb ram, quad core i7.  I've done most of my work in Ruby, so have done things using bundler, so I can only assume that this isn't normal for npm.  I wonder what's going on.  

After getting gulp to work I ran the normal build task and a TON of changes occurred, which surprised me.  I'm out of time for now, maybe I'll come back and hack on this some other day.

CyborgMaster:
Ok.  So I fixed the test and resolved the conflicts, but I found several other bugs that needed fixing.  So I started a new pull request (#2115).  I also narrowed the scope of the timeout.  Now the bare minimum of the link function is inside the timeout instead of the whole function.

PR #1560, opened 2015-03-10
Reset view to default if model date changes

blaind:
I'm using datepicker for functionality where users use a shared physical device for creating user accounts: datepicker is used for entering persons birthday. Problem is: when $scope.bday resets, datepicker selection window still has the previous bday active (and hence visible for next person - bad).

With this commit date picker view is selected to default, and $scope.$date is being reset. 

Tests pass. There was no references for testing e.g. $scope.$mode, so left tests untouched for now.

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -84,6 +84,9 @@ angular.module('mgcrea.ngStrap.datepicker', [
           if(angular.isDate(date) && !isNaN(date.getTime())) {
             $datepicker.$date = date;
             $picker.update.call($picker, date);
+          } else {
+            $datepicker.$date = date;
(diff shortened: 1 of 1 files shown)


mgcrea:
Could you please provide a plunkr as a text case? The fix might work but I would not expect it to live in the update function.

blaind:
I used update function just because it seemed to work there. Don't really know the internals of angular-strap that well :) 

Here is a plunker: http://embed.plnkr.co/sLtWYjCOHuDYyQ5OkSFX/preview 
Way to reproduce:
- select some date (e.g. year 2000)
- click save -> obj.date is being reset
- open date picker view, you see the previously selected date as preselected

PR #1563, opened 2015-03-11
Support `data-` prefix for ng-bind reference

monokrome:


mgcrea:
Should be harmonized with the tooltip.

monokrome:
@mgcrea Sorry, I think that I may be missing some context here?

juliusosokinas:
Is there an issue opened I could refer to? I'm having the same problem, when data-ng-bind does not work as expected.

monokrome:
@mgcrea What's up?

DaleWebb:
I think he meant that the change should also be made to the tooltip component

PR #1565, opened 2015-03-11
Add support for prefixEvent on bsModal

### LINKED ISSUE #1559: data-prefix-event is not used for bs-modal
The documentation says

> If provided, prefixes the events '.hide.before' '.hide' '.show.before' and '.show' with the passed in value. With the default value these events are 'modal.hide.before' 'modal.hide' 'modal.show.before' and 'modal.show'

But no matter what I specify for data-prefix-event on the html element, the event name is always modal.*, e.g. modal.show.before

Either the code needs to be updated, or the documentation :smile:

Tobbe:
prefixEvent was not a recognized option for bsModal.  Added it to list
of options, together with unit tests to make sure it works.

Closes #1559

### DIFF
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -300,7 +300,7 @@ angular.module('mgcrea.ngStrap.modal', ['mgcrea.ngStrap.helpers.dimensions'])
 
         // Directive options
         var options = {scope: scope, element: element, show: false};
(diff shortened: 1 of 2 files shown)


Tobbe:
Is there anything else I need to do to get this merged?

mgcrea:
Fixed by d781c14.

PR #1576, opened 2015-03-14
Don't stop event if no option is selected.

### LINKED ISSUE #1575: Error when pressing error with unfocused typeahead
If you disable typeahead autoSelect with the newly available option, you are able to press enter without selecting an option first. This throws an error in the `$typeahead.select` function, because the passed index is -1.

mzgajner:
This fixes #1575, which I reported.

New behaviour: when enter is pressed with the typeahead dropdown open but no option selected (because autoSelect is set to false), the situation is handled same as if the dropdown wasn't open.


ansonkao:
Awesome!!

mgcrea:
Thanks! Landed in e732244.

Please beware commit message guidelines for future patches.

PR #1581, opened 2015-03-18
Fix #1566 Incorrect tooltip positioning with 'manual' trigger

### LINKED ISSUE #1566: Incorrect tooltip positioning with 'manual' trigger
Here is the [plunk](http://plnkr.co/edit/1jmPmLOo4uvDYOpyHkNW?p=preview)

drew-r:
$timeout to allow digest cycle to complete if in progress. 
When using manual trigger the positioning is applied before the tooltip content is loaded.  The height of the tooltip is measured inaccurately and therefore the tooltip positioning is out.

This causes some of your tests to fail though - they are expecting positioning and rAF callbacks to be set / called sooner than they are as a result of the $timeout.

### DIFF
--- a/src/tooltip/tooltip.js
+++ b/src/tooltip/tooltip.js
@@ -222,34 +222,36 @@ angular.module('mgcrea.ngStrap.tooltip', ['mgcrea.ngStrap.helpers.dimensions'])
 
           $tooltip.$isShown = scope.$isShown = true;
           safeDigest(scope);
+          
+          // Allow $digest to complete if in progress
+          $timeout(function() {
+              // Now, apply placement
+              $tooltip.$applyPlacement();
 
-          // Now, apply placement
-          $tooltip.$applyPlacement();
+              // Once placed, animate it.
+              // Support v1.3+ $animate
+              // https://github.com/angular/angular.js/commit/bf0f5502b1bbfddc5cdd2f138efd9188b8c652a9
+              var promise = $animate.enter(tipElement, parent, after, enterAnimateCallback);
+              if(promise && promise.then) promise.then(enterAnimateCallback);
+              safeDigest(scope);
 
-          // Once placed, animate it.
-          // Support v1.3+ $animate
-          // https://github.com/angular/angular.js/commit/bf0f5502b1bbfddc5cdd2f138efd9188b8c652a9
-          var promise = $animate.enter(tipElement, parent, after, enterAnimateCallback);
-          if(promise && promise.then) promise.then(enterAnimateCallback);
-          safeDigest(scope);
+              $$rAF(function () {
+                // Once the tooltip is placed and the animation starts, make the tooltip visible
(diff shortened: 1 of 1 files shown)


mgcrea:
Thanks for the PR, however, `$timeout` is not really an option here, you can't trigger a full digest everytime you hover a tooltip, it would badly impact performance, please open an issue with a reproducible test case in a plunker.

drew-r:
You're welcome.
Did you see the plunker @albv put here
http://plnkr.co/edit/1jmPmLOo4uvDYOpyHkNW?p=preview

Also `safeDigest()` is called just prior to my edit, I thought this
triggered a full digest if one is not in progress anyway?
Thanks

michaelfdeberry:
@drew-r Using a setTimeout also seems to resolve the issue and won't trigger a digest.

drew-r:
@michaelfdeberry  thanks, confirmed. I wonder does this actually guarantee an in-progress $digest would complete before the callback is invoked?

michaelfdeberry:
@drew-r I was hoping someone else would chime in, but honestly I don't know enough about JavaScript or Angular to say it's guaranteed. 

However, since JavaScript is mostly single threaded and there is an inherent delay with setTimeout I believe it's a good chance that the $digest will finish first.

drew-r:
@mgcrea can you enlighten? 
I've pushed a commit to the above effect to drew-r:drew-r-patch-issue-#1566
I was hoping the pull request would reflect this but seems it doesn't. Perhaps it will if the request is reopened? Or I can submit as new pull request if preferable.
Commit is https://github.com/drew-r/angular-strap/commit/07a57497c588b550d960a64f9fd92714d4c00c79

mgcrea:
A safe digest is a local-scope-only digest that won't throw errors if there is another local/global digest occurring.
 
```
function safeDigest(scope) {
    scope.$$phase || (scope.$root && scope.$root.$$phase) || scope.$digest();
}
```

A `$timeout` on the other hand is basically a setTimeout followed by a `rootScope.$apply`, (see https://github.com/angular/angular.js/blob/v1.3.15/src/ng/timeout.js#L52)

Regarding the timing of setTimeout vs. an ongoing digest, since the digest cycle is a synchronous call. It won't yield control to the browser's event loop until it is done, (see https://github.com/angular/angular.js/blob/v1.3.15/src/ng/rootScope.js#L722), so you can be sure that any ongoing digest will end before the setTimeout is evaluated, (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/EventLoop).

evillemez:
For what it's worth, I copied the modified fix (using $window.setTimeout instead of $timeout) into my project, and it did fix the positioning problem.

https://github.com/drew-r/angular-strap/commit/07a57497c588b550d960a64f9fd92714d4c00c79

Any reason not to resubmit the PR with that fix?

PR #1584, opened 2015-03-19
add backdropAnimation to list of options

episodeyang:
this is to fix the same bug that appeared in `tooltip`, but in this case for modal.js 4f27655


mgcrea:
Can't merge this as is.

PR #1589, opened 2015-03-19
fix(typeahead): position the tooltip correctly even when it is not yet visible

bbloomf:
set the typeahead tooltip to force itself to be visible until it has been positioned, so that positioning will happen properly

closes issue 1588


ansonkao:
Awesome! I banged my head on the table for a while...
I really like that you found a fix that works, feels quite hacky though... is this the best way to fix it?

bbloomf:
I agree that it feels hacky and probably isn't the best way to fix it, but I thought it might be useful as a demonstration of the root cause of the bug.

ansonkao:
Hopefully someone familiar with the tooltip mechanics can chime in.

Issue #1417, opened 2015-01-08
Timepicker should support seconds and milliseconds in the dropdown


kkirsche:
Currently, the timepicker can only give an option to display a dropdown for hour, minutes, and AM/PM if it is 12 hour based (as I understand it). In the timeFormat option though, you offer the ability to set much more complex times such as `h:mm:ss.sss` which would include milliseconds. At the moment though, I haven't been able to find through the provided options how to easily support those more complex times for users through the dropdowns.

E.g.:
![screen shot 2015-01-08 at 2 48 24 pm](https://cloud.githubusercontent.com/assets/947110/5669434/6f8ab8da-9745-11e4-897f-2d38f481b4a4.png)

Am I missing something or is this something which would need to be added?

vmlf01:
Currently the timepicker dropdown only has support for manipulating hours/minutes.

honkskillet:
+1 for seconds.

honkskillet:
To elaborate, this would be very useful as an elements for selecting arbitrary amounts of time (instead of just times of the day).  5 minutes and 30 seconds, 30 hours and 30 minutes, 15 seconds, etc.    This could be used as a picker for a timer, for example.

mgcrea:
Landed in e1edb9a.

PR #1599, opened 2015-03-23
feat($timepicker): add support for seconds

kjaquier:
This makes possible to edit the seconds with the timepicker, as requested in #1417. I did not make the milliseconds but it should be pretty straighforward following my modifications.

I updated the unit tests so they also check the seconds, and added a few that are more specific to that feature. I edited the unit test for "roundDisplay" so it mocks the clock in order to be reproducible. I also updated the example in the documentation of the timepicker.

Potentially breaking changes :
* Seconds dropdown is shown automatically if timeFormat include them (ex : h:mm:ss)
* Added parameter secondStep
* Meridian field (AM/PM) is now at position 6 instead of 4 (replaced by seconds field)

PS : this is my first pull-request, I hope I did everything right but don't be too rude if it's not the case :sweat_smile: 

Edit : ok there are still bugs. Sorry about it, I'm gonna check that.

### DIFF
--- a/src/helpers/date-formatter.js
+++ b/src/helpers/date-formatter.js
@@ -24,7 +24,7 @@ angular.module('mgcrea.ngStrap.helpers.dateFormatter', [])
     };
 
     function splitTimeFormat(format) {
-      return /(h+)([:\.])?(m+)[ ]?(a?)/i.exec(format).slice(1);
+      return /(h+)([:\.])?(m+)([:\.])?(s*)[ ]?(a?)/i.exec(format).slice(1);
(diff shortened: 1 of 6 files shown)


kjaquier:
Ok everything is clean and functional now. I fixed the remaining errors, added some more tests, and removed a stupid breaking change I made to roundDisplay that was due to a misunderstanding.
I ovewrote the commit in order to keep a clean history.

mgcrea:
Nice PR, well done ;-). Merged in e1edb9a.

PR #1614, opened 2015-03-31
feat(typeahead): if value attr exist use value instead of label when selected

zxdong262:
change behavior when select object like :
```javascript
 {
    value: 'some value'
    ,label: 'label html'
}
```
output `value` to the input element instead of label.


vmlf01:
In my opinion, this goes against the typeahead way of working with the ng-option/bs-options mechanism. You should use those attributes instead to specify what you want in ng-model/input display. Otherwise, you get a fixed behavior and lose flexibility.

zxdong262:
@vmlf01 you are right, I did not fully understand how typeahead works then.

Issue #1616, opened 2015-03-31
"iconLeft" and "iconRight" option of datepicker to be html instead of class


zxdong262:
it is a feature request.

I know it is rare: 

for users do not use :before icons, it is not possible to customize the right&left arrow for now.
use html instead will solve this problem.

hope it works like this:

``` html
<button tabindex="-1" type="button" class="btn btn-default pull-left" ng-click="$selectPane(-1)" ng-bind-html="$iconLeft">
</button>
```

and default option will be:

``` javascript
{
    iconLeft: '&laquo;'
    ,iconRight: '&raquo;'
}
```

PR #1619, opened 2015-04-01
feat(datepicker): use html instead of icon class

zxdong262:
to make datepicker left and right icon more flexible

https://github.com/mgcrea/angular-strap/issues/1616


zxdong262:
select and else should be changed together, this is not enough.

Issue #1578, opened 2015-03-17
prefixEvent not considered for typeahead


manikantag:
Hi, `prefixEvent` is being considered in typeahead code.

     .directive('bsTypeahead', ["$window", "$parse", "$q", "$typeahead", "$parseOptions",
           function($window, $parse, $q, $typeahead, $parseOptions) {

    var defaults = $typeahead.defaults;

    return {
      restrict: 'EAC',
      require: 'ngModel',
      link: function postLink(scope, element, attr, controller) {

        // Directive options
        var options = {scope: scope};
        angular.forEach(['placement', 'container', 'delay', 'trigger', 'keyboard', 'html', 'animation', 'template', 'filter', 'limit', 'minLength', 'watchOptions', 'selectMode', 'autoSelect', 'comparator', 'id'], function(key) {
          if(angular.isDefined(attr[key])) options[key] = attr[key];
        });

But while emiting event, prefixEvent is being used.

    scope.$emit(options.prefixEvent + '.select', value, index, $typeahead);

Any reason for that?

vmlf01:
Similar to #1559

EliSnow:
Also note it is not available in `Popovers` and possibly `Datepickers`, `Timepickers` and `Selects`, which all utilize `Tooltip`.

vmlf01:
Fixed by  d781c14 .

PR #1624, opened 2015-04-02
fix(typeahead): add the prefixEvent option to typeahead

jbcrawford:
add the prefixEvent option to typeahead so that it can be changed from the default value


vmlf01:
Fix for #1578 ?

jbcrawford:
Looks like it, sorry I didn't find it for reference

mgcrea:
Fixed by d781c14.

PR #1648, opened 2015-04-16
Enable prefixEvent and prefixClass on several directives

### LINKED ISSUE #1559: data-prefix-event is not used for bs-modal

EliSnow:
This adds the 'prefixEvent' and 'prefixClass' configurable options to the directives of the datepicker, modal, popover, select, timepicker and typeahead modules.

These changes should not break backwards compatability. In some of the directives 'prefixEvent' and 'prefixClass' were not in the module defaults. I have added them in **commented out** as uncommenting them will break backwards compatability. I propose in a future major version that these lines be uncommented as well as the naming of these defaults be standardized (in some cases the event prefix name starts with `$` other times not).

Fixes #1559 and #1578

### DIFF
--- a/src/datepicker/datepicker.js
+++ b/src/datepicker/datepicker.js
@@ -9,6 +9,8 @@ angular.module('mgcrea.ngStrap.datepicker', [
 
     var defaults = this.defaults = {
       animation: 'am-fade',
+      //uncommenting the following line will break backwards compatability
+      // prefixEvent: 'datepicker',
       prefixClass: 'datepicker',
       placement: 'bottom-left',
       template: 'datepicker/datepicker.tpl.html',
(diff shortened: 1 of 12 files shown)


EliSnow:
@mgcrea, I need to fix the commit message. Would you classify this as a feature or fix? I could see it going either way.

vmlf01:
This also replaces PR #1565, right?

EliSnow:
and #1624, I believe.

mgcrea:
Defaults commenting is a bit odd, can you add some details? Why sometimes comment nothing, sometimes `prefixEvent` default, sometime both `prefixEvent` and `prefixClass`?

EliSnow:
`prefixEvent` and `prefixClass` were only added to defaults commented out because adding it in uncommented would break backwards compatibility if anyone was relying on the inherited default of 'tooltip'. Perhaps, because the availability of the events are undocumented, it may be viewed that it does not break backwards compatibility, and they can be uncommented.

In cases where either `prefixEvent` or `prefixClass` is uncommented, it was already in the module's defaults and was not added in by this PR.

mgcrea:
Landed in d781c14. Thanks.

PR #1650, opened 2015-04-16
docs(collapse): clarify disallowToggle description

### LINKED ISSUE #1549: Collapse disableToggle option is ambiguous in the docs
The docs say something about disabling dblclick, but in the code is appears it's only function is to "keep from toggling" on "click", nothing to do with dblclick. Not only that, but even that isn't explained well, as what the user will see is the inability to "close" the panel, as it's wired to "open only" when disallowToggle is true.

vmlf01:
Closes #1549

### DIFF
--- a/src/collapse/docs/collapse.demo.html
+++ b/src/collapse/docs/collapse.demo.html
@@ -126,7 +126,7 @@ $scope.panels.activePanel = {{ panels.activePanel }};
           <td>boolean</td>
           <td>false</td>
           <td>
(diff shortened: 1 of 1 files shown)


vmlf01:
@dankahle do you think this is clearer regarding disallowToggle function?

dankahle:
That looks better, at least they know what they're gonna get and that it has nothing to do with dblclick.

mgcrea:
Thanks, landed in f8f1cbe.

Issue #1654, opened 2015-04-20
null value in model causes "TypeError: scope.$activeIndex.indexOf is not a function"


nicklasb:
If one initializes a drop down with a null value in the model, this error is thrown:

     TypeError: scope.$activeIndex.indexOf is not a function

It would appear that starting with a null value in the model should not be a problem, null is a quite common value.

Plunkr: http://plnkr.co/SNhTvz

nicklasb:
I made a pull request(#1656) that resolves the issue(ignore the first #1655). 
At least I would think that is the way is *supposed* to work.

vmlf01:
Nice one.

Have a look at AngularStrap contribuiting guidelines: https://github.com/mgcrea/angular-strap/blob/master/CONTRIBUTING.md. You should make changes to the files inside src/ folder, because the dist/ files are generated by the build process.

Also, I made a test for reproducing this error, have a look here: https://github.com/vmlf01/angular-strap/commit/6fb75800c60395b123c7dffb07dc13b650579508. Can include it in your PR?

nicklasb:
Oh, obviously..Sorry, was a bit tired yesterday. 
Made a new PR at https://github.com/mgcrea/angular-strap/pull/1657 that changes /src instead.

Yes, please do.

vmlf01:
Lol, looks like I also typed it wrong! I meant if you could include it in your PR branch so it goes together.

nicklasb:
Aha, OK. :-)
I have included it now.

nicklasb:
I *was* briefly thinking about how you should do that, by making a pull request to my branch or something. But there the thinking ended. :-)

nicklasb:
It worked! :-)

nicklasb:
It is actually *really* cool how well pull requests, and GIT, in general works. 
Considering a time even before Visual Source Safe, it borders on miraculous sometimes.

mgcrea:
Fixed by 670f967.

PR #1655, opened 2015-04-20
Handle initial null values for multiple

nicklasb:
This makes https://github.com/mgcrea/angular-strap/issues/1654 go away by initializing $activeIndex to an array.


nicklasb:
Closing this one, https://github.com/mgcrea/angular-strap/issues/1656 is better, I would think.

PR #1657, opened 2015-04-20
Correctly initialize null values for multiple selects

nicklasb:
This makes mgcrea#1654 go away.

(this PR correctly makes the change in the /src folder)


nicklasb:
Just voting for this being a part of next minor version. If a select is a part of a new item, for example, supporting null-values are important. 
As javascript supports null values, it feels like editors should as well.

Also, I have stuff riding on this fix. :-)

mgcrea:
Landed in 670f967. For future PRs, please use the proper commit message format (check the contributing guidelines).

nicklasb:
Will do.

PR #1664, opened 2015-04-27
Make safeDigest() absolutely safe

e-oz:
It will solve [issue](https://github.com/angular/angular.js/issues/11713) with ngAnimate and in general I think it's more safe.

### DIFF
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -270,7 +270,9 @@ angular.module('mgcrea.ngStrap.modal', ['mgcrea.ngStrap.helpers.dimensions'])
       // Helper functions
 
       function safeDigest(scope) {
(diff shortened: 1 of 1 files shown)


mgcrea:
Not sure about this one, I know the performance hit would be negligible, but it might make finding bugs (like this one) harder.

vmlf01:
Maybe at least log the error to the console?

jamm:
@vmlf01 often e2e tests requires console.log to be empty after test, so it would make them fail.

seldary:
So, after angular closed [the animate callback issue](https://github.com/angular/angular.js/issues/11713), I'm guessing angular-strap now doesn't support angular 1.4?

This is more to do with calling $animate than safeDigest...

e-oz:
Can't reproduce anymore.

PR #1669, opened 2015-04-28
fix(datepicker): fix selected day display when picking first day of the ...

### LINKED ISSUE #1665: January 1st Does Not Work in Date Picker
There's something magical about January 1st with the date picker.  If you open the date picker and choose a date on a different month, say February, and then go back to January and choose Jan. 1st, the date picker will not show it highlighted as the selected date.  It corrects itself if you toggle to another  month and back.

This was tested in Chrome 42.0.2311.90 on the angular-strap site - http://mgcrea.github.io/angular-strap/#/datepickers

vmlf01:
...month

Fixes #1665


mgcrea:
Landed in 247d16d. Thanks!

PR #1675, opened 2015-04-30
Fix for 1666

danielAtAditro:
Tabbing out  of select should not prevent event propagation

### DIFF
--- a/src/select/select.js
+++ b/src/select/select.js
@@ -183,8 +183,8 @@ angular.module('mgcrea.ngStrap.select', ['mgcrea.ngStrap.tooltip', 'mgcrea.ngStr
             return $select.hide();
           }
 
-		  evt.preventDefault();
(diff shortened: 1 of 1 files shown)


src/select/select.js
```diff
@@ -177,14 +177,15 @@ angular.module('mgcrea.ngStrap.select', ['mgcrea.ngStrap.tooltip', 'mgcrea.ngStr
...
-          evt.preventDefault();
-          evt.stopPropagation();
 
           // release focus on tab
           if (options.multiple && evt.keyCode === 9) {
             return $select.hide();
           }
 
+          evt.preventDefault();
+          evt.stopPropagation();
```
jukkasi: This fix is incorrect since tab won't work properly if options.multiple === false

mgcrea:
Landed in 1ec917d.

PR #1677, opened 2015-05-04
docs(datepicker): document timezone option for bs-datepicker

hsrobmln:
Added missing documentation for the timezone option for bs-datepicker


mgcrea:
Thanks! Landed in 17fbba4.

Issue #786, opened 2014-05-23
No max selected option for multiple select


Mattnmoore:
Is there something I'm missing? It might be extremely useful to include a max-selected option so you can limit how many items a user can select. I'm trying to hack around this to do it, but nothing is really working.

strizzaflex:
I'm running into the same issue. We want to have a multiselect with a high end limit. If this is something we would like to see in the project proper I can give it shot when I get back next week.

strizzaflex:
I have this working I modified the directive a bit. Changed the watch for model a bit. 

Old Directive Watch:
```javascript
      // Watch model for changes
        scope.$watch(attr.ngModel, function(newValue, oldValue) {
          // console.warn('scope.$watch(%s)', attr.ngModel, newValue, oldValue);
          select.$updateActiveIndex();
          controller.$render();
        }, true);
```

Modified Directive Watch: 
```javascript
       // Watch model for changes
        scope.$watch(attr.ngModel, function(newValue, oldValue) {
          // console.warn('scope.$watch(%s)', attr.ngModel, newValue, oldValue);
          if(attr.limit < newValue.length){
            for(var i = 0; i < newValue.length; i++){
               if(i >= oldValue.length) break;
               if(!angular.equals(newValue[i], oldValue[i])) break;
            }

            if(i === newValue.length){
              newValue.pop();
            }else{
              newValue.splice(i, 1);
            }
          }

          select.$updateActiveIndex();
          controller.$render();
        }, true);
```

Can anyone tell me if this is where they expect the change would be made? Should I be doing this in the $select provider instead (not sure I have old and new values there although)?

PR #1679, opened 2015-05-05
feat(select): Add ability to have a high end limit on multiselect.

strizzaflex:
add ability to have high end limit on a multislect. This allows the user
to choose more than one option but no more than the high end limit.
currently there is no ability to limit the number of selected items.

based on issue #786


gwin003:
What is the status of this PR?

Ciboulette:
I've tried using data-limit="3", it didn't work.. Do you know why ?

Issue #1497, opened 2015-02-10
reposition popover after changes to template html


songty:
How can I recalculate the popover/tooltip placement after making changes to the template? Here is a [plunker](http://plnkr.co/edit/PDvOXZ2Kgcec1vDyrUyN?p=preview) example.

bstruthers:
if you're programmatically creating the popover via the service, you should be able to call the `$applyPlacement` method to have it recalculate the positioning.

PR #1681, opened 2015-05-06
feat (tooltip): add "redraw" event listener

EliSnow:
Tooltip now listens for a "redraw" event which signals that the displayed
tooltip should be repositioned. Use cases: When the triggering DOM
element is repositioned while the tooltip is active, the "redraw" event can
be emitted/broadcast and the tooltip will remains adjacent the DOM element.
When the dimensions of a tooltip's template change, the "redraw" event can
be emitted/broadcast and the tooltip will remain centered next to the
triggering DOM element.

This is a possible solution to #1497.

### DIFF
--- a/src/tooltip/test/tooltip.spec.js
+++ b/src/tooltip/test/tooltip.spec.js
@@ -578,6 +578,42 @@ describe('tooltip', function() {
 
   });
 
+  describe('redraw event listener', function () {
+    it('should listen for redraw events', function() {
+      var myTooltip = $tooltip(sandboxEl, templates['default'].scope.tooltip);
+      var applyPlacement = spyOn(myTooltip, '$applyPlacement');
+      scope.$digest();
+
+      myTooltip.show();
+      $rootScope.$broadcast('tooltip.redraw');
+      myTooltip.hide();
+      $animate.triggerCallbacks();
+      $rootScope.$broadcast('tooltip.redraw');
+
+      // $applyPlacement should have been called twice: immediately after show
+      // and again after the first $broadcast. The second $broadcast should not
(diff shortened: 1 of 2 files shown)


mgcrea:
Nice PR, would you please patch the docs as well?

EliSnow:
I'll see what I can do

mgcrea:
Not sure that this event handler is the way to go in the end, currently there isn't a single "active" event listener in the code and I'm not sure I want to break this paradigm: events are used to listen for component changes, scope methods to act on it.

Would a scope tooltip method `scope.$redraw` work for you (you could always manually setup the event handler after that).

EliSnow:
Sorry for abandoning this for so long.

With your proposed scope method, how would that integrate in with my controller? Currently my use case is that I have an event listener which fires on window scroll to redraw two popups. That code (using the event listeners) currently looks like:

```js
  angular.element($window).bind("scroll", () => {
    $scope.$broadcast("timepicker.popup.redraw");
    $scope.$broadcast("hamburger.popup.redraw");
  });
```

EliSnow:
@mgcrea, I have looked over the code and there doesn't appear to be a way to access the directive's scope from the controller, AFAICT. This is what I was aiming to resolve in this small use case.

Many angular-strap directives have their own scope methods and perhaps a better path would be to come up with an architecture of how those methods can be accessed from parent scopes. Your [comment here](https://github.com/mgcrea/angular-strap/pull/117#issuecomment-17061211) has one possible solution, but that doesn't seem like the idiomatic angular solution, IMO.

Closing for now, but I am interested in your feedback on this issue.

PR #1696, opened 2015-05-18
feat(date-parser): change parsing of 'y' format option to support parsing of 2 and 4 year

jeremy-compeat:
declarations with the same format

I am trying to support the ability to type in either a date with two or four digits for the year in
 a date. This is possible by setting the 'y' format for the year. However, inputting a two digit
date with the 'y' format produces a date from 20th century instead of the 21st. The change
suggested would take any two digit year below or equal to fifty and return the 21st century
equivalent. e.g. inputting a year of 15 now produces 2015 instead of 1915.

This change will cause anyone relying on the behavior of the 'y' format to produce 1915 from an
input of 14 to break as it would now produce 2015.

### DIFF
--- a/src/helpers/date-parser.js
+++ b/src/helpers/date-parser.js
@@ -114,7 +114,7 @@ angular.module('mgcrea.ngStrap.helpers.dateParser', [])
         'M'     : function(value) { return this.setMonth(1 * value - 1); },
         'yyyy'  : proto.setFullYear,
         'yy'    : function(value) { return this.setFullYear(2000 + 1 * value); },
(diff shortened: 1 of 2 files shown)


mgcrea:
@jeremy-compeat, could you please add the relevant uni tests, thanks!

jeremy-compeat:
@mgcrea, added unit tests for the parser. Let me know if they are sufficient or if you're looking for something more.

mgcrea:
Thanks! Landed in b3c276697a99ea6c4c4f18361ffea653d9a0956e.

PR #1698, opened 2015-05-19
feat(bsTabs): allow to bind selected tab by name

just-boris:
It is useful sometimes to select tab by its name. This feature introducing support of `name` attribute on `bs-pane`:

```html
<div bs-active-pane="tabs.activeTab" bs-tabs>
    <div ng-repeat="tab in tabs" title="{{ tab.title }}" name="{{ tab.title }}" disabled="{{ tab.disabled }}" ng-bind="tab.content" bs-pane></div>
</div>
```
If you set names for your tabs, it will be used as `bs-active-pane` instead of number. Backward binding also works, you can set string into `tabs.activeTab` and tab with this name is exists, it will be activated.

This feature allows to make wonderful things, such as tabs switching via `location.hash` or activating tab by name without sense for tabs order.

### DIFF
--- a/src/tab/docs/tab.demo.html
+++ b/src/tab/docs/tab.demo.html
@@ -156,6 +156,14 @@ $scope.tabs.activeTab = {{ tabs.activeTab }};
             <p>Disable pane</p>
           </td>
         </tr>
+        <tr>
+          <td>name</td>
+          <td>string</td>
+          <td>''</td>
+          <td>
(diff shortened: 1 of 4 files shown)


just-boris:
Also, I have an updated demo for the docs page, but I am not sure that it should be commited

mgcrea:
@just-boris Thanks for the PR, docs should be updated/commited too! Thanks! Landed in 5e83951.

PR #1700, opened 2015-05-19
fix(collapse): keep original ng-model

antonshmakov:
Make active panels model changes properly propagate to outer scope

### DIFF
--- a/src/collapse/collapse.js
+++ b/src/collapse/collapse.js
@@ -68,7 +68,7 @@ angular.module('mgcrea.ngStrap.collapse', [])
       self.$targets.$active = !self.$options.startCollapsed ? [0] : [];
       self.$setActive = $scope.$setActive = function(value) {
         if(angular.isArray(value)) {
-          self.$targets.$active = angular.copy(value);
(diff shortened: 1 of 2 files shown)


src/collapse/collapse.js
```diff
@@ -68,7 +68,7 @@ angular.module('mgcrea.ngStrap.collapse', [])
       self.$targets.$active = !self.$options.startCollapsed ? [0] : [];
       self.$setActive = $scope.$setActive = function(value) {
         if(angular.isArray(value)) {
-          self.$targets.$active = angular.copy(value);
+          self.$targets.$active = value;
```
antonshmakov: Not quite sure why there was ``` angular.copy ``` at the first place, but it adds extra difficulties to have persistent set of open panels if there is intermediate isolated scope (e.g. page controller -> custom conditional directive -> bs-collapse: custom directive will have all the changes, but page controller won't w/o extra watcher/formatter in custom directive scope)

mgcrea:
Thanks! Landed in 432e7e2.

PR #1701, opened 2015-05-20
docs(bsTabs): add tab names into demo

just-boris:
Update demo to show changes made in https://github.com/mgcrea/angular-strap/pull/1698


mgcrea:
Thanks! Landed in 9560731.

PR #1703, opened 2015-05-20
fix(typeahead): if display value cannot be determined, use the model value

### LINKED ISSUE #1702: typeahead not allowing input to display model value if value is not in the list.
Looks like this issue was introduced by https://github.com/mgcrea/angular-strap/commit/5e0544e9dd6304330ed3da9c1e9d30583f832ce1#diff-fc7f312e9f6092388f3fd87246a8afbb 

If the displayValue is undefined, the formatter is returning `''`.  If the model is not in the list of options, the displayValue cannot be determined and the input's value is set to blank.

example plnkr: http://plnkr.co/edit/1PU5wn8bYnQJ0BMjw4WA?p=preview

Since a typeahead is often used as suggestions for an input field, I would still expect whatever value it's fed to display.

bstruthers:
If the element's model is not found in the typeahead's options, use the model as the element's value instead of an empty string. Fixes #1702

### DIFF
--- a/src/typeahead/test/typeahead.spec.js
+++ b/src/typeahead/test/typeahead.spec.js
@@ -274,6 +274,10 @@ describe('typeahead', function () {
       expect(elm.val()).toBe(jQuery('div').html(scope.states[0]).text().trim());
     });
 
+    it('should use the model value if the display value cannot be determined', function () {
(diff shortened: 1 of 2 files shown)


mgcrea:
Thanks! Landed in a7c97d3.

Issue #1241, opened 2014-11-06
Picker arrow behavior does nothing with initial value empty


tinyfly:
To build off of #603. 

If I start with an initial empty model value and have the behavior type set as picker nothing happens when I click a picker arrow.

**Current behavior:**
When I click a picker arrow first nothing happens. However if I click on the time and then click the arrows everything works just fine.

**Expected behavior:**
On first click of one of the picker arrows, it should select the time too.

[Here is and example plunker](http://plnkr.co/edit/xzDh03baXVn4yztrhXnS?p=preview)

daemon1981:
:+1:

lemoigne:
+1 I have the same problem @tinyfly.

PR #1709, opened 2015-05-27
fix(timepicker): arrows click were busted when arrow-behavior is "picker"

daemon1981:
Arrows click were busted when arrow-behavior is "picker"

from https://github.com/mgcrea/angular-strap/issues/1241


daemon1981:
What do you think about this fix ?

If ok, I can complete tests.

mgcrea:
Thanks! Landed in dce6a5cbb699dd4436b90147f6da28afed36d9d1.

PR #1712, opened 2015-05-28
fix(select): correctly sort selected items using source item index

### LINKED ISSUE #1707: Sort option on select directive buggy when more than ten items present
The sort option on the select directive is based on the line
```
if(options.sort) scope.$activeIndex.sort();
```
which sorts the scope.$activeIndex array by index. However, Array.prototype.sort() sorts according to each character's Unicode code point value by default, which means that 10, 11, 12, etc. come before 2. (See http://stackoverflow.com/questions/1063007/arr-sort-does-not-sort-integers-correctly)

Hence, this should be ordered numerically using a simple numeric compare function.
```
function sortNumber(a,b) {
    return a - b;
}
numArray.sort(sortNumber);
```

vmlf01:
Javascript array sort() uses alpha sort by default. This passes a function to the sort() call that will do a numeric sort instead, as suggested by @sumigoma.

Fixes #1707


mgcrea:
Thanks! Landed in f6c5cb8.

PR #1714, opened 2015-05-29
chore(build): set gulp default task to execute build

vmlf01:


mgcrea:
Thanks! Landed in 969f710.

PR #1719, opened 2015-06-04
DatePicker Docs - null timezone explanation

Icidis:
Updated DatePicker docs to be more intuitive as it seems as if UTC is always being used as it is the only option which is not the case. If left null it will use the local timezone.


mgcrea:
Thanks! Landed in f228010.

PR #1726, opened 2015-06-09
Chore (bower.json): Remove moot `version` property from bower.json

kkirsche:
Chore (bower.json): Remove moot `version` property from bower.json

Per bower/bower.json-spec@a325da3


mgcrea:
I think it's still useful to have the `version` available after a `bower install`, it's a safe and quick way to check the version you have downloaded, with no extra cost (I'm using `mversion` to bump versions). So I guess I'll keep it for now. Thanks for the PR.

Issue #1715, opened 2015-06-02
Aside leaks memory by not destroying its own scope when closed


jametsi:
I use the current version of angular-strap (2.2.2) with angular 1.3.15 and noticed that the aside's scope never gets destroyed when using data-template. 
This can be reproduced by using the default template with a controller that only has the listener for $destroy-event -> It never gets fired.

vmlf01:
Not sure what is your exact setup. I have created a plunker with an example following your description and the destroy is triggered when the aside DOM element is destroyed, as expected.
http://plnkr.co/edit/0sOUcGNVky6uUhdX4z1v?p=preview

jametsi:
In your example the $destroy-event doesn't get triggered when I close the aside, so if I open & close it multiple times, I have several instances of the controller in memory. If I click the destroy-button to manually trigger the $destroy-event, I get multiple alerts from those controllers.

I can of course force the destroy by replacing the close-button's ng-click="$hide()" with ng-click="$hide(); $destroy();". 

So is this expected behavior? It surprised me, at least :)

vmlf01:
Ah, I see what you mean now, I didn't test the plunker enough.
The same thing happened with tooltip, but aside/modal is not based on tooltip, so this needs to have a similar fix applied to prevent the memory leak. Will try to come up with a PR in the next few days.
Thanks for the follow up.

lookfirst:
@vmlf01 close this?

vmlf01:
thanks.

PR #1727, opened 2015-06-09
Feature/fix modal aside scopes

vmlf01:
Not perfect but prevents modal/aside scopes from continuously leaking.

Maybe someone can come up with a better alternative solution/implementation?

#1715


mgcrea:
Thanks!

PR #1741, opened 2015-06-18
Correct element width if affix is middle

BaggersIO:
The `affix` class should be added **after** resolving the element width https://github.com/mgcrea/angular-strap/blob/master/src/affix/affix.js#L128

Otherwise `.affix` / `position: fixed` would alter the proper element width.

This bug occurs if you refresh the page when the affix is on middle.


schlotzz:
+1

mgcrea:
Thanks! Landed in cd30b67.

Issue #1752, opened 2015-06-25
typeahead trimValue option does not work with watchOptions 


daniel-nalbach:
Issue #1604 was resolved with PR #1439 and added the new trimValue option, but spaces are still removed if watchOptions is used. In fact, it looks like the mere presence of the watchOptions attribute whether true or false is enough to cause space trimming.

daniel-nalbach:
Ignore #1753, not the issue. Canceled it.

daniel-nalbach:
I set data-watch-options="false" on the bs-typeahead directive element and then added console.log statements to the typeahead file. The results are as follows.
```
options.watchOptions =  false (black in console not blue, meaning string)
typeof options.watchOptions =  string (expected boolean)
options.watchOptions === false?  false  (expected true)
```
This is a problem because the typeahead file uses:
```javascript
if (options.watchOptions)
```
That statement will always evaluate true since strings are truthy in JavaScript.

PR #1753, opened 2015-06-25
fix(typeahead): make watchOptions work with trimValue

daniel-nalbach:
Update to watchOptions code block to check for the trimValue option before trimming spaces from the parsed input. See issue #1752.


daniel-nalbach:
Bad assumption about code. This trim removes spaces from the option value, not the input value. Canceling pull request as I continue investigating the issue.

PR #1768, opened 2015-07-08
Fix for Chrome - v43.0.2357.130 doesn't consider autocomplete="off" but "false"

MathieuLorber:
Typeahead recently became impossible to use on Chrome, the browser doesn't respect autocomplete="off" in last version anymore, and the native autocomplete hides the typeahead... autocomplete="false" works

(see http://stackoverflow.com/questions/15738259/disabling-chrome-autofill)

### DIFF
--- a/src/typeahead/typeahead.js
+++ b/src/typeahead/typeahead.js
@@ -199,7 +199,7 @@ angular.module('mgcrea.ngStrap.typeahead', ['mgcrea.ngStrap.tooltip', 'mgcrea.ng
         });
 
         // Disable browser autocompletion
-        element.attr('autocomplete' ,'off');
+        element.attr('autocomplete' ,'false');
 
         // Build proper bsOptions
         var filter = options.filter || defaults.filter;


MathieuLorber:
I just rewrote the commit to respect the expected message format

mgcrea:
@MathieuLorber thanks for the PR, could you please remove the built files from the commit. I'll update them for the next release.

MathieuLorber:
Yes sorry, I've accidentally added them when renaming commit... It must be ok

mgcrea:
Thanks! Landed as 8d4e263.

MathieuLorber:
Thank you =)

MathieuLorber:
@mgcrea , this is very confusing, but Chrome 43.0.2357.134 now supports back 'off' and no more 'false'. This is the same with a canary version (46.0.2466.0). I'm searching for more information...

MathieuLorber:
The situation isn't clear : https://code.google.com/p/chromium/issues/detail?id=468153

But autocomplete="off" seems to actually work in current Chrome version - despite the last comment.

@mgcrea I think the most reasonable thing to do is to revert the PR...

Sorry about that.

MathieuLorber:
The Chromium issue is still getting new posts every week. With my current version, I have native autocomplete with 'off' again, and not with 'false'...

PR #1771, opened 2015-07-09
Added a "keep-end" attribute for bs-pane for tabs.

ballerabdude:
This attribute will allow a pane/tab to stay at the end and any new
tabs/panes added will be put in front of it.

Signed-off-by: Abdul Hagi Abdul.hagi@hotmail.com


ballerabdude:
Im not sure how I can create an example. Can someone help me? I did checkout gh-pages but I think I was going something wrong.

mgcrea:
I think the proper way to achieve that is with a custom template, eg:

```html
<ul class="nav" ng-class="$navClass" role="tablist">
  <li role="presentation" ng-repeat="$pane in $panes track by $index" ng-class="[ $isActive($pane, $index) ? $activeClass : '', $pane.disabled ? 'disabled' : '' ]">
    <a role="tab" data-toggle="tab" ng-click="!$pane.disabled && $setActive($pane.name || $index)" data-index="{{ $index }}" ng-bind-html="$pane.title" aria-controls="$pane.title"></a>
  </li>
  <li role="presentation">
    <a role="tab" data-toggle="tab">Add Pane</a>
  </li>
</ul>
<div ng-transclude class="tab-content">
</div>
```

ballerabdude:
@mgcrea I added a test in my first commit. I tested which tab would be at the end and the one with the keep-end attribute was at the end. keep-end was on the first tab. I can pull the latest code and resubmit if you like.

PR #1778, opened 2015-07-14
Reset typeahead 'selection' through keyboard navigation

vladgurovich:
Currently if you use arrows to select a match, there is no way to "unselect" since you can only set them to values between `0`(first match) and `$matches.length-1`(last match).

This change will allow you to reset your selection if you keyboard navigate higher than the first match -- which is a state where you start out before using up/down arrows. It does so by allowing $activeIndex range from `-1`(which is a default index when no selection took place) to `$matches.length-1`

I verified that no tests were broken


mgcrea:
Wouldn't listening for the `ESC` be more natural to drop the current selection?
Being able to loop through a long list could be a feature.

PR #1786, opened 2015-07-18
fix(datepicker, timepicker): respect readonly and disabled attributes (fixes #946)

### LINKED ISSUE #946: datePicker and timePicker show calendar when ng-readonly="true"
as the title says, the datePicker and timePicker both show the calendar controls even if they are using ng-readonly="true".  And upon selecting a value, it is actually set in the textbox which uses ng-model.

I believe the calendars should not disable at all.  The inputs are correctly not allowing a user to type which is correct

hermansje:


mgcrea:
Landed in 79235e9429f4b4a1ebec1540defbced7a5eabd3d. Thanks!

just-boris:
I am not sure that disable datepicker on `readonly` inputs is fine.
I use datepicker on that inputs to avoid wrong input from user and it is not working since `2.3.1`

PR #1794, opened 2015-07-23
chore(modal): remove double attribute check

hermansje:


mgcrea:
Thanks! Landed in 5e22884.

PR #1796, opened 2015-07-23
bstimepicker: Force ability to update model for custom form validation

mitcht:
This feature allows developers to specify their own validation by essentially disabling the internal validation. It still passes through an instanceof Date when the bstimepicker control is used, but allows $watch to work in consuming controllers so one can offer manual validation of times such as is found in the following code: http://hastebin.com/ovocihuvip.vhdl

This change does not affect the normal operation of the bstimepicker directive.  It is purely a method to disable internal validation when a value is typed into the form.


mitcht:
I'm going to create a branch on my fork so i can submit this pull request without putting it into my master

PR #1800, opened 2015-07-24
feat (bstimepicker / bsdatepicker): add change allowing support for typing / return viewValue for custom validation

mitcht:
This feature allows developers to specify their own validation by essentially disabling the internal validation. It still passes through an instanceof Date when the bstimepicker control is used, but allows $watch to work in consuming controllers so one can offer manual validation of times such as is found in the following code: http://hastebin.com/ovocihuvip.vhdl

This change does not affect the normal operation of the bstimepicker directive. It is purely a method to disable internal validation when a value is typed into the form.

### DIFF
--- a/src/timepicker/timepicker.js
+++ b/src/timepicker/timepicker.js
@@ -507,7 +507,7 @@ angular.module('mgcrea.ngStrap.timepicker', ['mgcrea.ngStrap.helpers.dateParser'
 
           // If the model needs to be updated to matter what, these validation routines
           // can be overridden by setting the attribute overrideValidation eq true
-          if (defaults.overrideValidation === 'true'){
+          if (defaults.overrideValidation === 'true' && viewValue instanceof Date === false){
             return viewValue;
           } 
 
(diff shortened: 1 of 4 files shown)


src/timepicker/timepicker.js
```diff
@@ -503,6 +504,13 @@ angular.module('mgcrea.ngStrap.timepicker', ['mgcrea.ngStrap.helpers.dateParser'
         controller.$parsers.unshift(function(viewValue) {
           // console.warn('$parser("%s"): viewValue=%o', element.attr('ng-model'), viewValue);
           var date;
+
+          // If the model needs to be updated to matter what, these validation routines
+          // can be overridden by setting the attribute overrideValidation eq true
+          if (defaults.overrideValidation === 'true'){
```
mitcht: All this does is force 'whatever was typed' to go back into the view model. That way it can be validated outside of the timepicker control.  In case for instance, i use 24 hr time format, or want to support seconds.  I can use momentjs and custom validation to ensure that a proper datetime object is set. Works similar to this section, but allows custom abilities.

mgcrea:
Could you please squash all your commits into a single one respecting the commit message format defined in the guidelines?

mitcht:
There, I think that *might* be correct.

mitcht:
Added validation override to datepicker.

mitcht:
@mgcrea Please review. Thx

mitcht:
@mgcrea bump.

mgcrea:
Thanks for the PR, looks good to me, just need unit-tests and cleanup of commits (all squashed into one).

PR #1802, opened 2015-07-24
Collapse all open

cdumas22:


mitcht:
if you broadcast on rootscope that this is occurring, you can also notify the page that it is occurring so 'other custom things' can also respond and close.

PR #1803, opened 2015-07-27
docs: affix requires debounce to be loaded

lookfirst:


mgcrea:
Thanks, landed in 3596d84.

lookfirst:
Thanks for an amazing library. =)

PR #1806, opened 2015-07-27
(datepicker): fix for IE 10 rendering of datepicker month selection mode

### LINKED ISSUE #1801: IE 10 month selection mode rendering
When viewing datepicker in month mode the options do not stretch across the whole picker in IE 10
http://plnkr.co/edit/pv2hqk1Ar5GEppjZ3zAy?p=preview

month-selection-ie10

# Steps to reproduce
1. Open up datepicker
2. Click month/year at top of the picker

# Solution

In datepicker.tpl.html change `ng-show="showLabels"` to `ng-if="showLabels"`

ckniffen:
Switched rending of week labels to use `ng-if` instead of `ng-show` because IE 10 was incorrectly factoring in the hidden table row when calculating `table-layout: fixed`

Fixes #1801

### DIFF
--- a/src/datepicker/datepicker.tpl.html
+++ b/src/datepicker/datepicker.tpl.html
@@ -18,7 +18,7 @@
         </button>
       </th>
     </tr>
-    <tr ng-show="showLabels" ng-bind-html="labels"></tr>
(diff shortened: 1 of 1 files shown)


mgcrea:
Thanks! Landed in 77003b2.

PR #1816, opened 2015-08-03
fix(select): when model is cleared, update active index and render accordingly

### LINKED ISSUE #1791: Clearing bs-select value
What is the proper way to clear bs-select (single-select) component value?

Setting $scope.select = {} works fine but does not set input invalid if required is used (and is not actually empty value).

Using $scope.select = null sets invalid fine but when select is opened the previous value is still displayed as selected in options which is incorrect. When select is opened second time, problem is gone.

EDIT: http://plnkr.co/edit/EnEOoUFmNw1HhXUbLanq

jukkasi:
Fixes #1791


mgcrea:
Thanks! Landed in 651fa7e.

PR #1818, opened 2015-08-11
Fix bug on select mutiple

polo2ro:
Fix bug if the selected item is not in the list of suggested values of a multiple suggest, add a test for this case.

### DIFF
--- a/src/select/select.js
+++ b/src/select/select.js
@@ -115,6 +115,9 @@ angular.module('mgcrea.ngStrap.select', ['mgcrea.ngStrap.tooltip', 'mgcrea.ngStr
             $select.activate(index);
             if(options.multiple) {
               controller.$setViewValue(scope.$activeIndex.map(function(index) {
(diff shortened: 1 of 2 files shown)


src/select/select.js
```diff
@@ -115,6 +115,9 @@ angular.module('mgcrea.ngStrap.select', ['mgcrea.ngStrap.tooltip', 'mgcrea.ngStr
             $select.activate(index);
             if(options.multiple) {
               controller.$setViewValue(scope.$activeIndex.map(function(index) {
+                if (undefined === scope.$matches[index]) {
```
mgcrea: Please use `angular.isUndefined()` here.

mgcrea:
Thanks for the PR! Looks good! Could you please squash your commits into a single one, using a commit message that do respect the contributing guide?

polo2ro:
reworked version in #1826

PR #1820, opened 2015-08-11
Fix for click event not updating the value

bryantlikes:
I was having an issue with the example on the website using the latest versions of angular-strap and bootstrap. The click event is not fired on the input since it is hidden by the button. The fix listens for the event on the active element which in this case would be the button.


mgcrea:
Can you please provide a reproducible test case?

PR #1823, opened 2015-08-13
fix(typeahead): don't add autocomplete if already present

bstruthers:
If the element already has an autocomplete attribute specified, there's
no need to add it or override it.


mgcrea:
Thanks! Landed in c73b684.

PR #1826, opened 2015-08-14
Fix bug on select mutiple

polo2ro:
Fix bug if the selected item is not in the list of suggested values of a multiple suggest, add a test for this case.
this is a reworked version of pull request #1818


mgcrea:
Thanks! Landed in a5e6518.

PR #1829, opened 2015-08-20
fix(typeahead): add attribute autocomplete="off" instead of autocomplete="false"

### LINKED ISSUE #1828: Typeahead autocomplete="false" should be autocomplete="off"
In Chrome I get browser autofill for typeahead because typeahead adds attribute autocomplete with value "false" instead of "off". After changing value to "off" everything works fine.

According to W3C the value should be "off"

http://www.w3.org/TR/2011/WD-html5-20110525/common-input-element-attributes.html#the-autocomplete-attribute

Firefox agrees it should be "off"

https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion

IE agrees it should be "off"

https://msdn.microsoft.com/en-us/library/ms533486(v=vs.85).aspx

jukkasi:
Fixes #1828

### DIFF
--- a/src/typeahead/test/typeahead.spec.js
+++ b/src/typeahead/test/typeahead.spec.js
@@ -300,7 +300,7 @@ describe('typeahead', function () {
 
     it('should add the autocomplete attribute if one is not already present', function () {
       var elm = compileDirective('default');
-      expect(elm.attr('autocomplete')).toBe('false');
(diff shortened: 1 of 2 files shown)


mgcrea:
Thanks for the PR, landed in eb48cca.

PR #1843, opened 2015-08-27
$typeahead. Added checking of whether was added keydown handler to element already or not

Niaro:
Good day mr. Mgcrea!
I'd like to ask you consider this pull request. In some cases there was a moment when
```js
$typeahead.show()
```
fires twice, and in the result we got two keydown handler on element, and that ruins logic of selecting by arrows.

Thank you for your attention and time spent on considering.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: node_js
 sudo: false
 node_js:
-  - "0.10"
+  - "stable"
 
 env:
   global:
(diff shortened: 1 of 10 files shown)


mgcrea:
Please provide a test-case.

jukkasi:
I have seen this issue too but it is hard to reproduce.

However this pull request does not seem to fix the root cause but only work around it.

PR #1849, opened 2015-08-31
Added async delay option

logorrhea:
Here's my stab at adding a delay to making async requests with the typeahead.


mgcrea:
Should already be handled by 2d3eb295ab23034e2f5838a2c68674d404d69028.

PR #1853, opened 2015-09-03
fix(select): fix clicking scrollbar and scrolling on IE

### LINKED ISSUE #1202: bs-select IE11 - clicking on scroler hides dropdown
If I apply custom css to limit the height of the list of options, it does not let me scroll in IE 11.
The following css is applied:
```
[bs-select] {
    /*position: relative;*/
    overflow: hidden;
    text-align: left;
}

.select.dropdown-menu {
    max-height: 100px;
    overflow-y: auto;
    /*width: 90%;*/
}
```

See [plunker example](http://plnkr.co/edit/17w963fz62PWdHYjmjCO?p=preview)

jukkasi:
Fixes #1202

### DIFF
--- a/src/select/select.js
+++ b/src/select/select.js
@@ -210,6 +210,19 @@ angular.module('mgcrea.ngStrap.select', ['mgcrea.ngStrap.tooltip', 'mgcrea.ngStr
           }
         };
 
+        $select.$isIE = function() {
+          var ua = $window.navigator.userAgent;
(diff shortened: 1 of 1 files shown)


mgcrea:
Thanks for the PR, however, I'm not fond of the additional `$utils` file if it is only used once, and has tied logic methods like `selectScrollFix`. Would you mind refactor this as private methods inside the `select.js` file?

jukkasi:
Done, hope it is OK now

mgcrea:
Thanks, landed in c280612. If I need the IE test somewhere, I'll reconsider having a separate utils module.

PR #1858, opened 2015-09-05
fix(select): Let event propagate when tab key is pressed

### LINKED ISSUE #1557: bs-select prevents correct work of Tab key
This [line](https://github.com/mgcrea/angular-strap/blob/master/src/select/select.js#L184) of code prevents correct work of Tab key and tabindex at all. When focus is set on bs-select subsequent press on Tab key will move focus somewhere but not where it should move.

jukkasi:
Fixes #1557


jukkasi:
Seems this is already in https://github.com/mgcrea/angular-strap/pull/1675

mgcrea:
Thanks for the PR! Landed in 1ec917d.

PR #1859, opened 2015-09-07
The list of options here is missing data-backdrop-animation

episodeyang:
this fixed the problem of not being able to set the backdrop animation for the modal-backdrop element. 

This also allows custom styling on the backdrop element.


episodeyang:
How does this PR look? It is a single line change so I'm not sure how come there are so many failed check. It would be great if we can merge!

PR #1860, opened 2015-09-08
docs: update modal to reflect resolve/locals support, fixes #1845

### LINKED ISSUE #1845: Support resolve method for Modal controllers
I didn't see anything mentioned in the docs about this so sorry if it is already supported... Now that https://github.com/mgcrea/angular-strap/issues/728 has been resolved(which is awesome btw!) it would be nice to be able to specify a resolve property for the controller. Similar to how ui-router functions...

```
$modal({
  controllerAs: 'myCtrl',
  controller: 'MyCtrl',
  resolve: {
    dependency: function() {
      return data;
    }
  }
});
```

Without this there are only two ways to pass data into the controller that I can see. 1. attach things to the scope 2. inline the controller function so it has access to the same context as the function instantiating the modal.

I don't like #1 because we are trying to avoid attaching things $scope whenever possible. #2 is not great because it makes it ugly to test the controller. We typically put our controllers, services, etc. in their own file and require them in as needed - it would be nice to be able to test the controller in isolation.

ryanray:
https://github.com/mgcrea/angular-strap/commit/b272203f7704d2f66873d974beb20e4ac9fd4e7f added support for two awesome(IMO) features that aren't documented. `resolve` and `locals` allow you to pass data into your modal's controller so you can avoid passing things around on `$scope`. This simple PR adds those two definitions to the modal docs.

### DIFF
--- a/src/modal/docs/modal.demo.html
+++ b/src/modal/docs/modal.demo.html
@@ -180,6 +180,22 @@
           </td>
         </tr>
         <tr>
+          <td>resolve</td>
+          <td>object</td>
+          <td>false</td>
+          <td>
+            <p>Object containing dependencies that will be injected into the controller's constructor when all the dependencies have resolved. The controller won't load if the promise is rejected.</p>
+          </td>
+        </tr>
+        <tr>
+          <td>locals</td>
+          <td>object</td>
+          <td>false</td>
+          <td>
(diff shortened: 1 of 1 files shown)


mgcrea:
Thanks! Landed in 4655109.

PR #1867, opened 2015-09-15
fix(typehead): Render properly when only one object is matched

### LINKED ISSUE #1866: bs-typeahead does not render object correctly
See plunker for details http://plnkr.co/edit/NvYt2PT5XCYydId7DrXW?p=preview

The problematic line is 
``
if (values.length === 1 && values[0].value === newValue) return;
``
which is located in typeahead component scope.$watch(attr.ngModel) function

jukkasi:
Fixes #1866


mgcrea:
Thanks! Landed in 4506d0e.

PR #1884, opened 2015-09-24
fix($tooltip): close popovers on escape key

### LINKED ISSUE #1537: Close on Esc doesn't work for popovers
Hi,

As I understand from the docs, escape key should close popovers when keyboard option set to `true`.
I put together this plunkr (edited from docs) http://plnkr.co/edit/UbPXUdY3vCNyOdEjYw1A to try and use this option, and it doesn't work.
I set the keyboard to `true` but esc key doesn't close popover. The problem is that the popover never gets focus (tested `document.activeElement`) so doesn't catch the `keyup` event to close the popover.

idanen:
Fixes #1537

### DIFF
--- a/src/popover/popover.tpl.html
+++ b/src/popover/popover.tpl.html
@@ -1,4 +1,4 @@
-<div class="popover">
+<div class="popover" tabindex="-1">
   <div class="arrow"></div>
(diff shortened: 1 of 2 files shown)


mgcrea:
Thanks! Landed in 97203b2.

Issue #521, opened 2014-03-04
Compatibility with UI Bootstrap


jontayesp:
If I load the angular-strap module alongside the "ui.bootstrap" module, several javascript errors are reported and components do not work. Ideally, it would be great to mix and match the best components from both libraries.

The UI Bootstrap library can be found here: http://angular-ui.github.io/bootstrap/

ClearCloud8:
+1

siulung4073:
+1

ryancat:
+1

grepnull:
+1

Florelli:
Same as @jsanders here !

Restricting myself to version 2.1.6 of angular-strap and including only the modules of bootstrap.ui that I'm using solved the issue.

gaui:
+1. I'm using Angular-Strap in conjunction with UI-Bootstrap and having conflicts between $modal. Was there some solution to this?

ryancat:
@gaui We are having same issue here. For now there is no clean way to resolve this as angular is resolving injectable providers using a internal key-value pair based on names. Our temporarily solution is using a static angular ui bootstrap library, and duplicate the ```$modal``` provider with a new name ```$uimodal```. As long as angular strap loads after angular ui bootstrap, you can always use ```$modal``` for angular strap, and ```$uimodal``` for angular ui bootstrap. 

It's essentially not a good practice to use both library as they are not promising compatibility.

gaui:
@ryancat Not a viable solution. This makes it hard to upgrade ui-bootstrap.

All directives are prefixed with `bs` - why not factories/services also? So it would be `$bsModal` or something.

ryancat:
@gaui Even with bs prefix, conflicts may happen in the future of angular ui bootstrap. Angular team may also use bs prefix in their next version of library (I have seen on their github page but cannot find it now).

gaui:
@ryancat Yes but there is an obvious conflict currently. The `bs` prefix was just a suggestion, because they have it in their directives. It can be `as` or whatever, I don't care. I just don't want Angular-Strap to conflict with ui-bootstrap. :)

ryancat:
haha sure -- I was working on unifying our third party libraries used in our front end codebase. The ```modal``` provider is just one of that. We were using old version of angular strap and manually changed some code for date picker. Now it's impossible to upgrade it without regression, sigh...

gaui:
Hehe, that's exactly what I'm trying to avoid. :smile:

[...]

PR #1885, opened 2015-09-24
chore(gulp): rename all providers, add `compat` task to `build`

frapontillo:
Rename all providers in `compat` task.
Providers are automatically renamed from `whatever` to `bsWhatever` and from `$something` to `$bsSomething` in order to keep consistency.

The `compat` task has also been added to the series of tasks executed before `build` so that the releaser will always release a compat build alongside the main one.

Addresses #521


mgcrea:
Landed in d063b44. Released in `v2.3.3`.

Thanks!

PR #1887, opened 2015-09-25
fix(gulp): do NOT rename $$rAF

frapontillo:
I was probably drunk. :beer:


mgcrea:
Landed in 847d938. Thanks!

PR #1891, opened 2015-09-27
fix (dropdown): empty dropdown

### LINKED ISSUE #1890: dropdown: don't show dropdown element if it's empty
Hello,

I use dropdown directive in a dynamic menu where some of its element haven't submenu. Because it's dynamic, the template contains a bsDropdown directive for each element.

Unfortunately, the dropdown template is append for all elements in the menu, on trigger event.

See the [JS Fiddle example](http://jsfiddle.net/79a75zxe/).

I would like to don't show dropdown element if it's empty.

grrseguin:
Fixes #1890


mgcrea:
Thanks! Landed in 4e38b56.

Issue #1668, opened 2015-04-28
Datepicker field doesn't update correctly when ng-model-options has updateOn value


arteme:
Consider the following code:

``` html
<form ng-model-options="{ updateOn: 'blur' }">
    <input type="text" ng-model="model.text" name="text">
    <input type="text" ng-model="model.date" name="date" bs-datepicker>
</form>
```

When the form has no ng-model-options, then `model.date` scope variable is updated when the date is selected from the drop-down and `date` input field is updated too.

When the form has ng-model-options with `updateOn` value, then `model.date` scope variable is updated when the date is selected from the drop-down, but `date` input field is updated to reflect the previously selected value.

Here is a live example: http://plnkr.co/edit/Pbsvo1HpP0lHQPGHtWZQ?p=preview

Similar issue has been observed by someone else with a different `updateOn` value: http://stackoverflow.com/questions/27377085/angular-strap-datepicker-updateon-submit

joshrickert:
+1

Removed my updateOn configuration for now and it is behaving correctly again.

arteme:
I have an ugly hack for a work-around when using the form wuth `updateOn: 'blur'`. I have a watch on the model and in there I force the field re-render as:
```javascript
$('[name=date]').scope().ngModel.$render();
```

DavidGiffordJr:
I was able to correct this by modifying controller.$render of datepicker.js by setting $dateValue to $viewValue as I was seeing these were sometimes not the same.

 ```javascript
controller.$render = function() {
    // console.warn('$render("%s"): viewValue=%o', element.attr('ng-model'), controller.$viewValue);
    if (angular.isDate(controller.$viewValue)) { controller.$dateValue = controller.$viewValue; }
    element.val(getDateFormattedString());
};
```

gwin003:
I just ran into this exact issue last Friday. Any ETA on a merge?

strizzaflex:
+1 For a merge.

pgorley:
@DavidGiffordJr's code resolved the issue for me.  Is there any chance we could see this merged?

nhartner:
+1 for merge. I wasted a lot of time trying to figure out why the datepicker input field would show a stale value. And the workaround isn't pretty.

nhartner:
Cool. Apparently the devs don't care about a blatant bug that's been open for 2 years with a PR submitted with test cases.

PR #1901, opened 2015-10-02
fix(datepicker): updateOn blur fix

DavidGiffordJr:
Datepicker does not update displayed date to selected when using updateOn: 'blur' ngModelOptions setting

GitHub issue #1668


mgcrea:
Please provide a test-case

PR #1906, opened 2015-10-09
add phantomjs as dependency

just-boris:
NPM v3 does not install peer dependencies by default. `phantomjs` is a peer dependency of `karma-phantomjs-launcher`. SO, I can't run tests with NPM v3.

You can choose any version of it, but it should be mentioned into your package.json now.

Furthermore, see https://github.com/ng-tools/undertaker-lib-tasks/pull/1 this is another issue which doesn't allow me to build `angular-strap` with NPM v3

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "karma-phantomjs-launcher": "^0.2.1",
     "karma-sauce-launcher": "^0.2.14",
     "ng-factory": "^1.0",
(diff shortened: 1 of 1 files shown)


just-boris:
Also I tested build on node.js v4, it works fine except this.

mgcrea:
Landed in 0c44d47c69a0d67c4d3d1e78ac07dd25701bbd1a.

PR #1907, opened 2015-10-09
fix(select): highlight null value option

just-boris:
I have a select with null option in my project. When this option is selected it is not highlighted as expected.
You should only check that value is defined, but doesn't need to check that value is actually truthy


mgcrea:
Thanks! Landed in 4dc6171.

PR #1911, opened 2015-10-14
[Doc]: add missed templateUrl to options

leftstick:
Since i found it was incorrect that `template can be provided as cache id`, it should the `templateUrl` option


mgcrea:
You should edit source files for docs:

`src/dropdown/docs/dropdown.demo.html`

PR #1921, opened 2015-10-21
feat(tooltip): Added exotic positions left-top, left-bottom, right-to…

lskramarov:
Hello, i'm added new exotic positions and test for it. In the near future I plan to put the automatic positioning for it and rework some code.


mgcrea:
Thanks! Landed in 2e53756.

PR #1927, opened 2015-10-27
Fix(bsSelect): bsSelect should rerender when values inside arrays changes

### LINKED ISSUE #1833: bs-select bs-options do not update when updating the $scope bs-options source option more than once
Having a problem figuring out why one can't update the displayed values in the bs-options more than once. You can edit one of the properties of a bs-option to be different, then edit the same bs-option again and the bs-option's value does not get updated in the display, but does get updated in the $scope. 

Please see the plunker:
http://run.plnkr.co/plunks/Ca69CM/

nickknissen:
Change bsSelect to deep watch bsOptions values.

Didn't know if this is a fix or a feature. So i decided to call it a fix. 
Will gladly rebase.

fixes #1833


mgcrea:
Thanks! Landed in ee9177d.

PR #1930, opened 2015-10-28
fix opening dropdown with custom template

js-padavan:
Hello,
after #1891  dropdowns could not be opened without content property on scope. However, custom templates will probably have another variables to bind, instead of $scope.content.


mgcrea:
Fixed by https://github.com/mgcrea/angular-strap/commit/89f9608f1c7578689d098b9b592f85b1dd52e492#diff-8b5edbc424c9072182bb9410189dbe20L63

PR #1938, opened 2015-11-02
feat: Add active state to the dropdown items

eugef:
Add class "active" to the dropdown item if property "active" is true.


mgcrea:
Thanks! Landed in fac071e.

Issue #1897, opened 2015-10-01
fullDate format not working in Spanish


overloadnn:
Datepicker does not work when using spanish locale. The problem is with the "d" character in the fullDate entry in the locale file.

For example: angular-locale_es-ar.js  --> **fullDate: "EEEE, d 'de' MMMM 'de' y"**

In $dateParser.parse the valid date **martes, 13 de octubre de 2015** cannot be resolved with the regex **/^(domingo|lunes|martes|miércoles|jueves|viernes|sábado), (0?[1-9]|[1-2][0-9]|3[01]) '(0?[1-9]|[1-2][0-9]|3[01])e' (enero|febrero|marzo|abril|mayo|junio|julio|agosto|septiembre|octubre|noviembre|diciembre) '(0?[1-9]|[1-2][0-9]|3[01])e' (-?0*[0-9]{1,4})$/i**

sdoran:
I'm seeing the same issue for Spanish and Portuguese. The regex generated by date-parser.js doesn't take into account 'de'. 

I've reproduced the issue in Plunkr http://plnkr.co/edit/oY8jiGXXGTnKUAaUiGFC?p=preview

vmlf01:
It looks like the $dateParser is blindly replacing the ```d``` token with the day parsing regex when it should ignore the ```'de'``` token because it is escaped inside the quotes. 

In fact, $dateParser should ignore all literals inside the format string. I'll have a look and see it I can come up with a quick fix.

sdoran:
That would be awesome thanks Victor. Even a workaround would do but I've come up short trying to find one.

PR #1939, opened 2015-11-02
Fix date parser literals

vmlf01:
This will be a fix for #1897.

However, I'm having some trouble coming up with a regular expression to cope with escaped quotes inside the literals, i.e. ```3 '0''clock'```, so it's still not finished.

Help anyone?

### DIFF
--- a/src/helpers/date-parser.js
+++ b/src/helpers/date-parser.js
@@ -228,29 +228,6 @@ angular.module('mgcrea.ngStrap.helpers.dateParser', [])
 
       // Private functions
 
-      function setMapForFormat(format) {
-        var keys = Object.keys(setFnMap), i;
-        var map = [], sortedMap = [];
-        // Map to setFn
-        var clonedFormat = format;
-        for(i = 0; i < keys.length; i++) {
-          if(format.split(keys[i]).length > 1) {
-            var index = clonedFormat.search(keys[i]);
-            format = format.split(keys[i]).join('');
-            if(setFnMap[keys[i]]) {
-              map[index] = setFnMap[keys[i]];
-            }
-          }
-        }
-        // Sort result map
-        angular.forEach(map, function(v) {
-          // conditional required since angular.forEach broke around v1.2.21
-          // related pr: https://github.com/angular/angular.js/pull/8525
-          if(v) sortedMap.push(v);
-        });
-        return sortedMap;
-      }
-
       function regExpForFormat(format) {
         // `format` string can contain literal values.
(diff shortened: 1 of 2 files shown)


mgcrea:
You mean `3 '0\'clock'`? Or actually `3 '0''clock'`? Is a double single quote an escape?

mgcrea:
[This patch](https://github.com/mgcrea/angular-strap/commit/7a1fd530daeed410e6c462472ce2bd9c7f9fbe12) seems to pass the tests, what do you think?

vmlf01:
I mean the second case, from angular date filter (https://github.com/angular/angular.js/blob/master/src/ng/filter/filters.js#L402), they are using two single quotes as an escape sequence, instead of a back-slash.

But your patch gave me another idea, maybe we can replace two consecutive single quotes with the back-slash escape sequence at the beginning and then use the regex for escaped quotes to match it.
That might be easier to understand than handle the empty formatPart tokens.

vmlf01:
Changed last commit to support the escaped literals, but now it seems that the date value parsing in https://github.com/mgcrea/angular-strap/blob/master/src/helpers/date-parser.js#L231 also has the original problem.

It seems very similar to the regex build code just fixed, so maybe it is possible to re-use it. I'll continue to check tomorrow.

vmlf01:
I think I fixed everything now, but I still have a question about line https://github.com/vmlf01/angular-strap/blob/77cb19cf48735f717334e7c1b27ce6a81d4e9f50/src/helpers/date-parser.js#L281

Before it was
```
        format = escapeReservedSymbols(format);
```
and I changed it to 
```
        re = escapeReservedSymbols(re);
```
because ```format``` is a local variable, so that line did nothing, and I thought it was just a typing error, but in reality, I don't know what the escapeReservedSymbols purpose is or if it is needed.

mgcrea:
Hum. Looks like a good fix, it looks like `escapeReservedSymbols` is supposed to escape symbols that have special meaning in a regex, for instance, a matches any character, but if you want to specifically match a dot, you have to `replace(/\./g, '[.]')`.

mgcrea:
Hum having a second look at it, I don't think it's suppose to be used there, as it will break matched patterns. It should probably be applied before on the format, we need extra tests to check this.

belfz:
Has this been released yet? I think that the problem referenced in #1897 still occurs even with version 2.3.5

vmlf01:
No it hasn't, it will be in next version.

PR #1946, opened 2015-11-05
feat(modal): set incremental z-index value for modals and asides to allow for stacking

DaleWebb:
feat(modal): set incremental z-index value for modals and asides to allow for stacking

The number of active modals and asides are tracked in order to set a `z-index` value that will allow those elements to be on top of each other.
Currently, if two asides are shown at the same time, clicking the backdrop will hide the aside that is below the one above.  Also, the backdrop of the second aside is below the first aside.

### DIFF
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -197,8 +197,10 @@ angular.module('mgcrea.ngStrap.modal', ['mgcrea.ngStrap.core', 'mgcrea.ngStrap.h
         $modal.hide = function() {
           if(!$modal.$isShown) return;
 
-          // decrement number of modals
-          backdropCount--;
+          if(options.backdrop) {
+              // decrement number of modals
+              backdropCount--;
+          }
 
           if(scope.$emit(options.prefixEvent + '.hide.before', $modal).defaultPrevented) {
             return;
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -28,8 +28,9 @@ angular.module('mgcrea.ngStrap.modal', ['mgcrea.ngStrap.core', 'mgcrea.ngStrap.h
       var requestAnimationFrame = $window.requestAnimationFrame || $window.setTimeout;
       var bodyElement = angular.element($window.document.body);
 
-      var modalCount = 0;
-      var baseZindexValue = 1037;
+      var backdropCount = 0;
+      var dialogBaseZindex = 1050;
+      var backdropBaseZindex = 1040;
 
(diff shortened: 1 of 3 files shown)


src/modal/modal.js
```diff
@@ -28,6 +28,9 @@ angular.module('mgcrea.ngStrap.modal', ['mgcrea.ngStrap.core', 'mgcrea.ngStrap.h
       var requestAnimationFrame = $window.requestAnimationFrame || $window.setTimeout;
       var bodyElement = angular.element($window.document.body);
 
+      var modalCount = 0;
+      var baseZindexValue = 1037;
```
mgcrea: Any reason you picked this specific number? Looks like it should default to 1050 according to https://github.com/twbs/bootstrap/blob/master/less/variables.less#L277

DaleWebb: I used the default value for the backdrop set here previously, which was
1038. Then took advantage of the fact that the element is one of the latter
child elements of the body to state that it is on top.

Should I set the modal z-index to start at 1050?

mgcrea:
We should closely follow bootstrap variables to avoid breaking people apps. Also we should manage separate z-indexes for backdrop and modal, first one being 1040/1050 (same as bootstrap), second one 1060/1070, etc. So that you still can easily manage items in-between layers.

DaleWebb:
Okay, so first modal/aside has the backdrop set at 1040 and dialog set at
1050, then the second backdrop being 1060 and dialog being 1070 as so on?

mgcrea:
Yep.

DaleWebb:
I've added unit tests about the z-indexes for each `.aside` and `.modal` and adjusted the starting value of the `z-index` to follow the bootstrap convention.  In a previous commit, this pull-request affected the other components that inherited `$modal` e.g. `$alert`, so now the `z-index` modifications will only happen if the component has a backdrop.

src/modal/modal.js
```diff
@@ -184,6 +197,9 @@ angular.module('mgcrea.ngStrap.modal', ['mgcrea.ngStrap.core', 'mgcrea.ngStrap.h
         $modal.hide = function() {
           if(!$modal.$isShown) return;
 
+          // decrement number of modals
+          backdropCount--;
```
mgcrea: should be wrapped in a `if(options.backdrop)`.

DaleWebb:
Updated

mgcrea:
Thanks! Merged :+1:

Issue #1941, opened 2015-11-03
dropdown directive bug


cyberdemon8:
i've found a bug on bsDropdown directive on line 2873 of dist/angular-strap.js
https://github.com/mgcrea/angular-strap/blob/master/dist/angular-strap.js#L2873

modified from:
```javascript
options.scope = scope;
```
to
```javascript
if (!options) {
  options = {};
}
options.scope = scope;
```
without this correction in certain cases (dropdown in a modal) angular rises this error:
```
TypeError: Cannot set property 'scope' of null
    at postLink (http://localhost:3000/bower_components/angular-strap/dist/angular-strap.js:2873:25)
    at invokeLinkFn (http://localhost:3000/bower_components/angular/angular.js:8281:9)
    at nodeLinkFn (http://localhost:3000/bower_components/angular/angular.js:7791:11)
    at compositeLinkFn (http://localhost:3000/bower_components/angular/angular.js:7140:13)
    at compositeLinkFn (http://localhost:3000/bower_components/angular/angular.js:7143:13)
    at compositeLinkFn (http://localhost:3000/bower_components/angular/angular.js:7143:13)
    at compositeLinkFn (http://localhost:3000/bower_components/angular/angular.js:7143:13)
    at compositeLinkFn (http://localhost:3000/bower_components/angular/angular.js:7143:13)
    at nodeLinkFn (http://localhost:3000/bower_components/angular/angular.js:7786:24)
    at compositeLinkFn (http://localhost:3000/bower_components/angular/angular.js:7140:13)
```

edoarsla:
Having the same issue with this one.  I have the top-of-tree angular-strap.

Plunker:
http://plnkr.co/edit/DvonnSFhxHuKbXupvGbS?p=info

cyberdemon8:
i found a better fix for this directive:
```javascript
directive('bsDropdown', [ '$window', '$sce', '$dropdown', function($window, $sce, $dropdown) {
    return {
      restrict: 'EAC',
      scope: true,
      compile: function(tElement, tAttrs) {
        //removed declaration of var options
        if (!tAttrs.bsDropdown) {
          var nextSibling = tElement[0].nextSibling;
          while (nextSibling && nextSibling.nodeType !== 1) {
            nextSibling = nextSibling.nextSibling;
          }
          if (nextSibling.classList.contains('dropdown-menu')) {
            tAttrs.template = nextSibling.outerHTML; //replaced options with tAttrs
[...]

PR #1947, opened 2015-11-05
fix(dropdown): options are null if dropdown with inline sibling template  is rendered again after it was previously destroyed

eugef:
if dropdown has inline sibling template then template is saved to the options during the compile time.
After dropdown is destroyed (for example it was inside the modal) and rendered again (modal reopened) - compile for dropdown is not executed and options are null.
This causes error "angular TypeError: Cannot set property 'scope' of null" and dropdown loses its template.

Solution is to save the template not in the options but in the attributes that are not destroyed.

### DIFF
--- a/src/dropdown/dropdown.js
+++ b/src/dropdown/dropdown.js
@@ -109,9 +109,6 @@ angular.module('mgcrea.ngStrap.dropdown', ['mgcrea.ngStrap.tooltip'])
       scope: true,
       compile: function(tElement, tAttrs) {
 
-        // Directive options
-        var options = {};
-
         // Support for inlined template (next sibling)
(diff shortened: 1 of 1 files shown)


eugef:
After in-depth testing I can confirm that this error "angular TypeError: Cannot set property 'scope' of null" exists for all dropdowns, not only with inline sibling template.

mgcrea:
Any chance you could write an unit test to edge against the error.

vmlf01:
Looks like this will be a fix for #1941

mgcrea:
Thanks! Landed in 8f2d1bc.

eugef:
@mgcrea, any chances this fix will be released soon?

PR #1948, opened 2015-11-07
fix(dropdown): ng-repeat problem with replaced array

keithbranton:
Dropdown breaks inside ng-repeat whenever the array driving the ng-repeat gets replaced with a new one. Compile only gets called once in this case, but scope.$destroy gets called whenever the rows get deleted and replaced by ng-repeat - options gets set to null, and all subsequent calls to postlink log an error. 

This fix finds the topmost scope containing $index, and uses its parent. If none is found then the current scope is used. A $destroy listener is attached to that scope (at most once) to set options to null.


vmlf01:
Seems to be solving the same thing as #1947.

keithbranton:
So it is - closing... Thanks.

Issue #1567, opened 2015-03-11
Typeahead promise bug in Angular 1.4.x Error: [filter:notarray] 


trevcor:
I know that 1.4 is still in beta but I wanted to cal this out.  When passing a promise to bs-options angular throws the below error:

    Error: [filter:notarray] http://errors.angularjs.org/1.4.0-beta.5/filter/notarray?    p0=%7B%22%24%24state%22%3A%7B%22status%22%3A0%7D%7D

jamm:
Caused by this breaking change: https://github.com/angular/angular.js/issues/10885
It's only bug I have with 1.4 and I can't switch dependency...

trevcor:
That bug cites IE9.  Sorry I didn't mention my browser, was using Chrome Version 41.0.2272.89 (64-bit).  So did they make a fix for IE9 that caused this bug?

jamm:
@trevcor sorry, I just linked wrong issue (many tabs opened).
Now fixed.
Thanks, @justincy, we love you.

yaboi:
bump - having this issue too and can't switch dependencies away from Angular beta...

mrweiland:
Is this on the road to fix - i cannot update to angular 1.4 either because of this. .:(

capaj:
@mrweiland you can try my fix, but I guess it is going to take a bit more time to fix the failing tests

Hrafnkellos:
We're having this problem as well, with Angular-Strap 2.2.2 and AngularJS 1.4.1

Line 1719 in angular-strap.js

	$parseOptions.valuesFn = function(scope, controller) {
	  return $q.when(valuesFn(scope, controller)).then(function(values) {
		$parseOptions.$values = values ? parseValues(values, scope) : {};
		return $parseOptions.$values;
	  });
	};

Then AngularJS calls `$parseOptions.valuesFn` and here, it crashes.

Line 18220 in angular.js

	if (!isArrayLike(array)) {
	  if (array == null) {
		return array;
	  } else {
		throw minErr('filter')('notarray', 'Expected array but received: {0}', array);
	  }
	}

Here `array` is a Promise object, not an array. So we get the following error:

    Error: [filter:notarray] Expected array but received: {}

Is this a issue in Angular-Strap or in AngularJS 1.4.x ?

christrude:
@capaj Hey man, you said try your fix but I can't seem to find a fix you do for this?

Bump for fixing this asap.

christrude:
[...]

PR #1958, opened 2015-11-17
Angular 1.4.7 fix filter:notarray

lukaszpaczos:
Connected with this issue:
https://github.com/mgcrea/angular-strap/issues/1567

When I switch from angular 1.3.15 to 1.4.7 i saw that i got error like this:

Error: [filter:notarray] Expected array but received: false

So i propose to add this simple fix to avoid it.

Best regards

### DIFF
--- a/dist/angular-strap.js
+++ b/dist/angular-strap.js
@@ -708,6 +708,9 @@
           return $filter('filter')(results, expression, comparator);
         });
       } else {
+        if (typeof array !== 'array') {
+          array = [];
+        }
         return $filter('filter')(array, expression, comparator);
       }
     };


vmlf01:
Do you have a sample with the problem you could share? Why would the array variable not be an array? Just by setting it to an empty array you might avoid the error but cause other side-effects (what was it set to before?).

lukaszpaczos:
You have right...I don't realize that I has "return false" instead of "return []" and that was the reason why I got error.

I fixed this because I thought was angular-strap problem, but in new version of angular (I changed from 1.3.15 to 1.4.7) they start to check in filter that the variable is array, so when on old version of angular everything working well I thought that maybe something wrong with angular-strap :)

So fix is not needed :)

Cheers

PR #1962, opened 2015-11-18
feat:(collapse): add support for the disabled attribute

gamelover42:
Add support to the bsCollapseToggle directive for the disabled attribute either via standard html or ngDisabled. This is to allow individual panels to be locked either open or closed when an action is required inside a panel. For example a form may need to be completed inside a panel.


mgcrea:
Thanks! Landed in 3d7fca4.

PR #1964, opened 2015-11-19
reversed calculation for top and bottom - fixed

eranbo:
When calculating the position for auto 'top' and auto 'bottom' of the $tooltip.$applyPlacement it looks that the code it dealing the opposite direction for top and bottom.
tested it on bsDropDown and it works good.

### DIFF
--- a/src/tooltip/tooltip.js
+++ b/src/tooltip/tooltip.js
@@ -357,10 +357,10 @@ angular.module('mgcrea.ngStrap.tooltip', ['mgcrea.ngStrap.core', 'mgcrea.ngStrap
             var originalPlacement = placement;
             var viewportPosition = getPosition($tooltip.$viewport);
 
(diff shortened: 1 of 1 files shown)


vmlf01:
Looks like this was changed back in commit https://github.com/eranbo/angular-strap/commit/01e900955cd360803c060bea36f5c2e954b2212a, but I find it strange that there are specific tests for these positions and they seem to pass with old code and new code, something has to be wrong with the tests/code

eranbo:
@vmlf01 
I see in tooltip.js on line 361 i the old cold the if clause is:
(originalPlacement.indexOf('bottom') >= 0 && elementPosition.bottom + tipHeight > viewportPosition.bottom

while in the new code the if clause is checking for top instead of bottom.

This causing the tooltip to get draw in the opposite direction.

vmlf01:
I'll try to have a look at the tests if I get some free time this weekend

Issue #1988, opened 2015-12-16
Datepickers/tooltips not respecting the "auto" placement attribute for top and bottom


jordancamara:
Only in 2.3.6, when using date pickers or other controls that use tooltips, setting their placement attribute to "auto" does not seem to automatically reorient the popover when close to the top or bottom of the screen. Looking into it, in lines 360..363 of tooltip.js it seems the logic for detecting the replacement of top/bottom was transposed in the most recent update.

jordancamara:
Fixed by #1989

PR #1989, opened 2015-12-16
Properly check the tooltip top and bottom placements.

jordancamara:
Hi, I noticed that the logic in tooltip that checks the auto placement for top and bottom seems to be transposed. Fixes issue #1988.


vmlf01:
Duplicate of #1964, didn't have time to look at the tests, maybe someone could have a look to check why they don't fail with this issue?

jordancamara:
I can look if I have time tomorrow. Do you want me to close this pull request since there is already one in place?

PR #1990, opened 2015-12-17
refactor(datepicker/select/timepicker): use character classes over ca…

toddbranch:
…pture groups for userAgent matching


mgcrea:
Thanks! Landed in 086d1ce.

toddbranch:
:+1:

PR #1992, opened 2015-12-17
feat(bsDatepicker): ngModelOptions.allowInvalid support

DenysVuika:
Added support for [ngModelOptions.allowInvalid](https://docs.angularjs.org/api/ng/directive/ngModelOptions) for cases when additional custom validation is performed at model level.


stevehipwell:
Any progress on this?

Issue #1561, opened 2015-03-10
Put ngModel on collapse is adding $dirty class to form element


omkadiri:
binding data using ngModel on a bs-collapse is adding $dirty class to the form element on the page

omkadiri:
This issue can prevent the use of $dirty for validation on a global scope. Eg '$scope.formName.$dirty' will return true even when the user have not interact with the form elements

joaoborsoi:
+1

PR #2002, opened 2015-12-30
Currently the collapse directive is using ngModel to bind the active …

joaoborsoi:
…panels index. This causes the collapse control to be included in the validation pipeline and when used inside a form, causes the form to be $dirty. (see issues #1561)

This commit deprecates usage of ngModel and adds a replacement bsActivePanels attribute with the same functionality, so it should be a matter of replacing ngModel references with bsActivePane. It is based on a similar solution made for the tabs directive (see issue #1233)

### DIFF
--- a/src/collapse/collapse.js
+++ b/src/collapse/collapse.js
@@ -32,7 +32,10 @@ angular.module('mgcrea.ngStrap.collapse', [])
       self.$toggles = [];
       self.$targets = [];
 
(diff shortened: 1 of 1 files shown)


mgcrea:
Interesting, but it needs docs & tests update. Also maybe we could refactor this `activePaneChangeListeners` logic as a helper service to deduplicate code with the bsTabs implementation.

PR #2005, opened 2016-01-04
docs(bsTabs): Change title to data-title in examples and tests

### LINKED ISSUE #2004: Tabs: bs-pane title html displayed as tooltip in some browsers
When mouse is positioned inside a tab with title attribute, some browsers display html source in the title as a tooltip.

Solution would be either to:
a) simply rename title attribute to data-title as is used in some other directives
b) refactor title to be a separate directive with transclusion + fallback to old title attribute if title directive is not present in the bs-pane. I prefer this option because tab headers 

Plunker: http://plnkr.co/edit/9KeRJtv3LwOLAyarFjSG?p=preview
Screenshot: selection_002

drola:
Change title to data-title in examples and tests. This is changed because browsers
display contents of title attribute when hovering over a tab in a tooltip. Existing
code already accepts data-title attribute. For that reason it was only necessary to
updated example code and tests.

Closes #2004.

### DIFF
--- a/src/tab/docs/tab.demo.html
+++ b/src/tab/docs/tab.demo.html
@@ -23,7 +23,7 @@ $scope.tabs.activeTab = "{{ tabs.activeTab }}";
     <hr />
     <!-- bsActivePane is optional -->
     <div bs-active-pane="tabs.activeTab" bs-tabs>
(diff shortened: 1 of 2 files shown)


mgcrea:
Thanks! Landed in 95d87ab.

PR #2009, opened 2016-01-07
feat(modal): support CSS class on modal element

### LINKED ISSUE #1331: modal: add custom class so that different modal on a single page can be styled differently
Basically what I am looking for is the ability to style both the backdrop element and the modal element via custom classes.

Any thoughts? 

and thanks for fixing the backdropAnimation bug.

rexm:
Add ‘modalClass' as an option on the modal configuration. When specified, the string will be added as a CSS class to the outer modal element. If the option isn’t specified, no class is added.

Closes #1331

### DIFF
--- a/src/modal/modal.js
+++ b/src/modal/modal.js
@@ -18,7 +18,8 @@ angular.module('mgcrea.ngStrap.modal', ['mgcrea.ngStrap.core', 'mgcrea.ngStrap.h
       backdrop: true,
       keyboard: true,
       html: false,
-      show: true
+      show: true,
+      modalClass: false
     };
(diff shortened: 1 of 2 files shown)


mgcrea:
Thanks for the PR!
Since modal is a base class for many components, I don't think we should use the term `modal` in the option key, at least in the service. I'm going to refactor this a little to use an `extraClass` option on the service instead (but aliasing it as `modalClass` in the directive).

mgcrea:
Landed in 0e9e3cd0afe0b9c9ac90b27345a8fd4dd72ef344

Issue #2007, opened 2016-01-07
bs-tooltip contentTemplate is ignored


cudasteve:
At version 2.3.6 (current) bs-tooltip appears to be completely ignoring the contentTemplate attribute.

#1306 sort of gets at a similar issue except back then the contentTemplate was working, just required the title.  [Here's a plunkr](http://plnkr.co/edit/2b1VTyM7AaN8S5FbKupM) I forked from that and just changed the version numbers for angularStrap.  As you can see, it's completely ignoring the contentTemplate.

Seems like it might have gotten completely removed [back in July](https://github.com/mgcrea/angular-strap/commit/623bfe0a94f34862c6a9df1eceeae6ebdba31862#diff-e8665eb306ddd26399e5b80f00dae5bbL90).  Perhaps that was unintentional since the commit seems to be about something else?

mgcrea:
It was not removed but refactored away in https://github.com/mgcrea/angular-strap/blob/master/src/helpers/compiler.js#L110

Not sure why it does not work, this should be tested.

cudasteve:
https://github.com/mgcrea/angular-strap/blob/master/src/tooltip/test/tooltip.spec.js#L912

cudasteve:
I think I figured it out.  The compiler is looking for `[ng-bind="content"]` when the tooltip would match `[ng-bind="title"]`
https://github.com/mgcrea/angular-strap/blob/master/src/helpers/compiler.js#L115

I'll submit a PR shortly though I'm not sure if you'd rather bring tooltip inline with the rest of the library by having it use a `content` attribute instead of a `title` one.

mgcrea:
Thanks! Landed in a9db388.

PR #2023, opened 2016-01-26
fix(popover): fix popover title (fixes #2022)

### LINKED ISSUE #2022: bs-popover title overload with content
using data-content-Template overload the title with "content" so we get content in title and content in title. In other words content show up twice. Here is plunker showing the problem

https://plnkr.co/edit/ALDwTUXJY22efsygWhb1

trampi:
The change in #2007 overrides all "title"-bindings with the contentTemplate. This fix renames the contentTemplate-attribute for bs-tooltip to titleTemplate, so that the name clash with the title of bs-popover is solved.

Breaking change: the contentTemplate-attribute for tooltips has been renamed to titleTemplate.

### DIFF
--- a/src/helpers/compiler.js
+++ b/src/helpers/compiler.js
@@ -106,13 +106,25 @@ function bsCompilerService($q, $http, $injector, $compile, $controller, $templat
       throw new Error('Missing `template` / `templateUrl` option.');
     }
 
+    if (options.titleTemplate) {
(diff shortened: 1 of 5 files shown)


mgcrea:
Thanks! Landed in 82b2ad2.

quonlee:
Listed changed are not present in the 2.3.7 build

PR #2027, opened 2016-01-29
Scrollspy: processing of $destroy is incorrect.

kaishuu0123:
Behavior of splice method:
  An array containing the deleted elements If only one element is removed, an array of one element is returned.

and splice method is destructive.

refs: JS splice method
https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/splice


mgcrea:
Thanks! Landed in e18ceaf.

PR #2028, opened 2016-01-30
Remove classList from dropdown to make compatible with IE9

cudasteve:
Theoretically fixes #9171, though I haven't run any tests


mgcrea:
Thanks landed in 7c9f1b5!

PR #2034, opened 2016-02-04
fix(tooltip): fix popover not closing when triggering element becomes disabled

### LINKED ISSUE #1905: Tooltip does not hide from meantime disabled object
### Please check the following steps

1. Move mouse to the firt button for show tooltip.
2. Pleas, click on the Button with tooltip button for disable it with ng-disable after the tooltip shown.
3. Move out mouse from the button, and the tooltip will never hide.

-

4. Make re-enabled the Button with tooltip button with the second one.
5. Move mouse to the first button again (noting will be happened).
6. However, when I moved out the mouse from the re-enabled button, the toolt will hide automatically.

### Demo environment
http://plnkr.co/edit/i24IJ3lIND390bqBz9Q5?p=preview

jchen86:
Fixes #1905

### DIFF
--- a/src/tooltip/test/tooltip.spec.js
+++ b/src/tooltip/test/tooltip.spec.js
@@ -200,6 +200,10 @@ describe('tooltip', function() {
     'bsTooltip-noValue': {
       scope: {title: 'Inherited Title'},
       element: '<a bs-tooltip>hover me</a>'
+    },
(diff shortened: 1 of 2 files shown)


mgcrea:
Thanks for the nice PR! Landed in a9b0040.

PR #2038, opened 2016-02-06
Added custom timezone

Danxil:


mgcrea:
Thanks for the PR! Looks good but it needs unit-tests!

manikantag:
Since AngularJS 1.4.x, `ngModelOptions` supprots giving custom timezone (in `+0530' format) for **date**, **time** input types. I tried using `bs-datepicker` with **date** input, like below:

```
<input type="date" 
	ng-model="selectedDate"
	ng-model-options="{ timezone:'+0530' }"
	data-date-format="yyyy-MM-dd"
	data-date-type="string" 
	bs-datepicker>
```

With `date-format="string"`, model is updated with selected date in UTC string format. But if I change it to `date-format="number"`, model is not updated with any value as AngualrJS expected it's value to be in String format adhering to UTC date string format (https://github.com/angular/angular.js/blob/v1.6.4/src/ng/directive/input.js#L1436). 

So, if all you need is the selected date in UTC string format, we don't need to support timezones explicitly in `bs-datepicker` (from Angular 1.4.x). But in my case, I need UTC millies and with `date-format="number"`, AngularJS is not allowing. 

Any solution other than converting the UTC date string into **Date** object and then using `Date.getTime()` to get millies?

Issue #1898, opened 2015-10-02
Problem with object binding, wenn I change focus and type in full name of selection


Cyber1000:
See this plunker: http://plnkr.co/edit/K9ymzdFjFoXxxdWOmlOb?p=preview
1) Type in the complete name, not only part of a state: Alabama
2) Press enter
3) Now the state (see on top of plunker) is $scope.selectedState = "{"id":1,"name":"Alabama"}";
4) Press tab to switch to the next input field
5) Suddenly the state changes to $scope.selectedState = "Alabama";

So I dont have the full object any more.

This works only, if I try this on a newly refreshed plunker page not on subsequent tries.
If I only enter part of the state (e.g. Ala) and press enter and tab I don't have this effect. Only if I write the full name.
Tested with last chrome and firefox.

Thanks

andras-tim:
At first, I can not find your problem. But when I reload the page and tried your scenario again I have got this too. Nice catch!

andras-tim:
@Cyber1000 This issue has a 0. step:
have to empty the input filed before write or past the complete name of state

andras-tim:
I spend a lot of time to debug this bug, but I can't find the root cause.

However **Ray Bateman** solved / workarounded this problem in mgcrea/angular-strap/#2061 pull request. I'm waiting for integrate this fix.

PR #2061, opened 2016-03-11
fix(typeahead): remove the ‘change’ event handlers from the typeahead element

### LINKED ISSUE #1319: Typeahed not working on Firefox when using objects in model
Typeahead does not work (on first selection) on Firefox when ng-model is using objects.
I'm using the examples in http://mgcrea.github.io/angular-strap/#/asides#typeaheads

Steps:
- Write "glob" in the second example (icons).
- Select "Globe" option.
- Click outside the input.

Actual result:
- The model transform from an object to the string "Globe"

Expected result:
- The model should remain an object.

Note:
This only happens the first time an option is selected.

rumbcam:
Stop ngModel from handling ‘change’ event fired by firefox on
input loss of focus to prevent an undesired change from ngModel to $viewValue.
Previously, on firefox, if $viewValue was an object set by typeahead,
this event would change $viewValue to the string value on the input element.

Could break other directives that rely on the ’change’ event if used on the same element as typeahead (fixes #1319, #652)

### DIFF
--- a/src/typeahead/test/typeahead.spec.js
+++ b/src/typeahead/test/typeahead.spec.js
@@ -173,10 +173,10 @@ describe('typeahead', function () {
       angular.element(elm[0]).triggerHandler('focus');
       elm.val(scope.states[0].substr(0, 2));
       expect(elm.val()).toBe(scope.states[0].substr(0, 2));
-      angular.element(elm[0]).triggerHandler('change');
+      angular.element(elm[0]).triggerHandler('input');
       elm.val(elm.val() + scope.states[0].substr(2, 4));
       expect(elm.val()).toBe(scope.states[0].substr(0, 2 + 4));
-      angular.element(elm[0]).triggerHandler('change');
+      angular.element(elm[0]).triggerHandler('input');
       expect(sandboxEl.find('.dropdown-menu li').length).toBe(1);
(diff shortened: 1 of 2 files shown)


andras-tim:
We have an other bug, what solved by this fix: #1898 This bug is independent of the Firefox, I have issue with it in the latest Chrome too.

andras-tim:
@mgcrea Sorry for my question, but are there any update for this issue? This fix is must have for me, therefore I must use a fork of your project while you integrate this. Thx!

mgcrea:
@andras-tim since it appears to be a breaking change, I was a bit reluctant to merge it. But if you confirm that it affects latest chrome too, I might merge it anyway.

mgcrea:
Landed in ebf6f98.

mgcrea:
Waiting for the build to pass.

andras-tim:
@mgcrea I use this patch since 12 days on latest Chrome with your several widgets. So, I hope, this will not breaks anything. - Thx for your fast answer & merges!

rumbcam:
The only way this will break anything is if someone were to use another directive with typeahead and they were trying to use the change event, but if they did that then it would have been pointless to be using typeahead anyway as they would then override anything that typeahead is doing.

Thinking through what happens this fix makes typahead work in a more expected way as it's value won't get changed later by ngModel after typeahead has already set it to what we expect it to be.

I am also currently using a forked branch with this fix

Issue #1754, opened 2015-06-26
Timepicker displays extra colon for HH:mm format


nikarh:
Timepicker unconditionally shows table column to display *timeSeparator* to separate minutes and seconds, even if *showSeconds* is set to false. This behavior could be observed in timepicker example on projects homepage — http://mgcrea.github.io/angular-strap/#/timepickers

![2015-06-26 10_45_30-angularstrap - angularjs 1 2 native directives for twitter bootstrap 3](https://cloud.githubusercontent.com/assets/5517011/8372916/860edd22-1bf0-11e5-9e25-02cbffd70401.png)

mfc-julius:
+1

george3447:
+1

phhien203:
+1

humpedli:
+1

mgcrea:
Fixed in e6f02bf.

PR #2064, opened 2016-03-15
Removed trailing colon in timepicker when seconds are not present

nikarh:
Fixed issue #1754


mgcrea:
Thanks for the PR, landed in e6f02bf. Beware of commit message guidelines.

Issue #2059, opened 2016-03-11
Typeahead, support for full text search


signalkuppe:
Maybe i'm doing something wrong, but I can't figure out how to search for plurals
If I search for "tables" I can see the matched results in the response from the server, but the typeahead doesn't display the results since the document found contains the word "table".
Typeahead seems to filter the results matching the $viewValue.
Is there a way to avoid this?

Thanks!

DaleWebb:
Hmm, this shouldn't be the case, as `bs-typeahead` basically just allows a dropdown to show results.  Whatever those results are is based on the array you return from `$http` call.

Could you show me the function that is meant to return the results and the HTML element that `bs-typeahead` is connected to?

DaleWebb:
This might be to do with the `$filter` that `bs-typeahead` has called `bsAsyncFilter`. https://github.com/mgcrea/angular-strap/blob/master/src/typeahead/typeahead.js#L186

DaleWebb:
@signalkuppe You could create a custom filter which just returns all of the results, then pass the name of that filter in through the `data-filter=""` attribute to your `bs-typeahead` element.  That should return all of your results.

signalkuppe:
Thanks Dale for your reply!
I'll give it a try.

PR #2074, opened 2016-03-31
feat(typeahead): allow disabling of a filter

### LINKED ISSUE #2058: Can typeahead display the result as partial match?
I'm having and issue with angular-strap typeahead as the plunker

http://plnkr.co/edit/v2BdeoqNFLJV4QRkXYxC?p=preview

In the Address (async via maps.googleapis.com) input text ,

I'm filled "bangkok, thailand" the result displaying correctly.

Anyway if I filled "Thailand bangkok" there're no result even in the console log return result correctly 

Can I edit the matching function or override somewhere to display all the returned result?

Thanks,

DaleWebb:
feat(typeahead): allow disabling of a filter to returned unfiltered results

The `bs-typeahead` directive will allow the `data-filter=""` attribute to take a
false value, which indicates that there should be no filtering of results.
Currently, the results are filtered by `bsAsyncFilter` by default, which may cause
undesired behaviour when retriving results from a remote source.

closes #2058 #2059


mgcrea:
Thank you for the PR! It's great that you added unit tests :+1: 
Just be careful with eslint errors as they break the build (check with `npm test`).

PR #2076, opened 2016-03-31
fix(typeahead): check if value is defined instead of boolean check

### LINKED ISSUE #2053: Typeahead does not accept value with id 0
Hi,

While working with typeahead I came across the issue that if it has an object with id 0, then it is not able to select that object if I am trying to set the model value from controller. I also tried it on plunker and it gave the same error. Due to this, the required validation is also failing on it. Following is the plunker that I have created:

http://plnkr.co/edit/l34zB6?p=preview

In this I modified icons list to have id's and initialized the model to id 0 by default. As you can see, it does not select the value in the list.

If I repeat the above process with id 1, then it works fine. Please find the update plunker with updated value below:

http://plnkr.co/edit/eqmBWy?p=preview

DaleWebb:
When the typeahead is first rendered, it checks if there is a `modelValue`
to show.  It does this by `if(modelValue && ...)`. if modelValue is false or 0,
it will not populate the typeahead display.  By checking if it is defined instead,
then it will accept 0 as a value.

closes #2053

### DIFF
--- a/src/typeahead/typeahead.js
+++ b/src/typeahead/typeahead.js
@@ -283,7 +283,7 @@ angular.module('mgcrea.ngStrap.typeahead', ['mgcrea.ngStrap.tooltip', 'mgcrea.ng
 
           // If there's no display value, attempt to use the modelValue.
           // If the model is an object not much we can do
-          if (modelValue && typeof modelValue !== 'object') {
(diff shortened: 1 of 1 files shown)


mgcrea:
Thanks! Landed with unit tests in c696fb6.

Issue #2052, opened 2016-02-29
bsSelecte & touchscreen


cp2587:
Hi,

We are using your library for almost a year and it works great so kudos for the great work. 
However, we just discovered what may be a bug: it seems that for touch screen (i have an xps15 with touchscreen) with google chrome, the following line in tooltip.js
```javascript
if (nodeName === 'button' && trigger !== 'hover') {
       element.on(isTouch ? 'touchstart' : 'mousedown', $tooltip.$onFocusElementMouseDown);
 }
```  
creates a weird behaviour in the bsSelect directive. When you click the button, it correctly shows the list but if you click it again, it does not hide the list. However if you are using the touch event it is working correctly. On a non-touchscreen computer, the behavior is the correct one too.
Would it be possible to bind both event : touchstart & mousedown to fix this ? Or maybe use 
```javascript
var isTouch = ('createTouch' in $window.document) && isNative;
```
like in select.js ? (in tooltip.js you do not check the isNative boolean)

DaleWebb:
Hi @cp2587 

I don't have a touchscreen computer, so I'm unable to reproduce this.  If you could do some hacking around and maybe submit a PR if you find the solution, that would be great.

What happens when you click off the `bs-select` (anywhere else on the page)?

Are you suggesting that the presence of the `isNative` value would possibly solve this?

cp2587:
Using the following lines in tooltip.js seem to resolve the bug (same as the ones in select.js):
```javascript
var isNative = /(ip[ao]d|iphone|android)/ig.test($window.navigator.userAgent);
var isTouch = ('createTouch' in $window.document) && isNative;
```

I made you a PR. On my side the fix worked fine for both touch & non-touch screen.

mgcrea:
Fixed by 6733ef4.

PR #2078, opened 2016-04-01
fix(select): fix popover not closing when click on select button on a touch screen

cp2587:
When you click the button, it correctly shows the list but if you click it again, it does not hide the list. However if you are using the touch event it is working correctly. On a non-touchscreen computer, the behavior is the correct one too.
Now the touch event is activated for mobile devices

closes: #2052


DaleWebb:
@mgcrea Is this okay to go ahead?

mgcrea:
Thanks! Landed in 6733ef4.

Issue #1083, opened 2014-09-17
Typeahead Keydown Issue on Webkit


jcq:
When editing existing text in a bs-typeahead control, the cursor jumps to the end of the text input after every keydown event.  This may be an issue localized to Webkit-based browsers, as I can reproduce it on Chrome and Safari, but not Firefox (tested both Mac and Windows).

While this isn't that much of a problem when entering text when we start with a blank field, it becomes a large problem when making edits to a field with existing text in it. 

This issue can be reproduced using the [default Typeahead examples](http://mgcrea.github.io/angular-strap/##typeaheads).

HarelM:
@vmlf01 Is this closed? I'm still getting it on chrome with latest version of angular and angular strap...?

HarelM:
Any one? any workarounds at least?

Kamilius:
I'm also getting this issue

DaleWebb:
Looks like there was confusion between two pull requests in the past that covered the issue, but it may not have been merged.

I'll see if I can reproduce the bug and take a look at what the fix was in those, then PR a fix (and maybe a temp solution).

Kamilius:
@DaleWebb thanks for info, Dale.

Kamilius:
@DaleWebb currently checking v2.3.6 of Angular-strap, and for me, issue still persists. I'm entering "sample text" into input, and when trying to `delete` ample, symbol by symbol, cursor jumps to the end of input, after first keypress.

DaleWebb:
Release `v2.3.8` doesn't contain the fix, however the `master` branch does

Kamilius:
Ok, thanks :)

Issue #1837, opened 2015-08-24
Typeahead - put cursor at start and type jumps cursor to the end


mattgoldspink:
Reproduction here:

http://plnkr.co/edit/7AFxRlQwfqaqTf9m45ql?p=preview

Looks like calling controller.$render() each time causes the value to be reset on the input and then the cursor jumps to the end. I'll submit the patch separately to maintain the cursor position

mgcrea:
Fixed by #2079

Issue #1448, opened 2015-01-21
Typeahead Moves to End of The Field When Typing


kkirsche:
The typeahead module doesn't respect where you are in an input and instead pulls you to the end of the input.

For example, if you type in the sample "rgi" and realize you you forgot the beginning, you can't jump to the first part of the input and add "Vi" to get "Virgi" as you type "Virginia"

Example:
https://www.youtube.com/watch?v=L7YdT8SSWm0

![typeahead](https://cloud.githubusercontent.com/assets/947110/5838195/9da21dfe-a153-11e4-8eb1-596ddad1db48.gif)
 
After further research this seems to be caused by this line in the typeahead module:
```javascript
element.val(selected ? selected.toString().replace(/<(?:.|\n)*?>/gm, '').trim() : '');
```

If we don't call `controller.$render();` in this block at the end this fixes this issue:

```javascript
// Watch model for changes
        scope.$watch(attr.ngModel, function(newValue, oldValue) {
          // console.warn('$watch', element.attr('ng-model'), newValue);
          scope.$modelValue = newValue; // Publish modelValue on scope for custom templates
          parsedOptions.valuesFn(scope, controller)
          .then(function(values) {
            // Prevent input with no future prospect if selectMode is truthy
            // @TODO test selectMode
            if(options.selectMode && !values.length && newValue.length > 0) {
              controller.$setViewValue(controller.$viewValue.substring(0, controller.$viewValue.length - 1));
              return;
            }
            if(values.length > limit) values = values.slice(0, limit);
            var isVisible = typeahead.$isVisible();
            isVisible && typeahead.update(values);
            // Do not re-queue an update if a correct value has been selected
            if(values.length === 1 && values[0].value === newValue) return;
            !isVisible && typeahead.update(values);
            // Queue a new rendering that will leverage collection loading
            controller.$render();
          });
        });
```

bstruthers:
Interestingly enough, this doesn't happen to me with a Firefox nightly but it does in Chrome.

kkirsche:
Interesting. Thanks for sharing that info @bstruthers

vmlf01:
Duplicate of #1083

meishier:
thanks. modify typeahead module, 
          var ss = element[0].selectionStart, sd = element[0].selectionEnd; //addd
          element.val(options.trimValue === false ? value : value.trim());
          element[0].setSelectionRange(ss, sd); //add
can fix it.

mgcrea:
[...]

PR #2079, opened 2016-04-01
fix(typeahead): stop the cursor from jumping to the end of the input upon refocus

DaleWebb:
currently, when finishing text input, blur then focus and begin typing at another point
in the text (e.g. the beginning), the cursor will move to the end of the text after
inserting a new character.
now, the cursor will move to after the new character typed (normal behaviour)

closes: #1083 #1837 #1448


mgcrea:
Thanks, I've added the missing tests in a separate commit.

PR #2081, opened 2016-04-05
Modify tab template to allow keyboard navigation.

RoryStokes:
Adding an empty href to the _a_ tag for tab headers allows keyboard navigation to the links (ie using tab) without changing functionality. Necessary to comply with WCAG 2.1 (https://www.w3.org/TR/UNDERSTANDING-WCAG20/keyboard-operation.html) and more in line with the w3c convention for hyperlinks (http://w3c.github.io/html-reference/a.html#a.attrs.href).


DaleWebb:
I think I've had issues before with combining `ng-click` with a `href`.  Where if you click the element, the `ng-click` will fire and so will the `href`.  Is there a way to use `data-tabindex` to get the same behaviour?

RoryStokes:
As long as the `href` attribute is empty it shouldn't cause any unwanted behaviour (often `href="#"` is used, which does fire). That being said, `tabindex="0"` should result in the same behaviour.

DaleWebb:
Just tried it out, works for me. :+1:

DaleWebb:
Okay for a merge, @mgcrea?

mgcrea:
Yep! Landed in af2fa21. Thanks!
PS: Beware of commit message guidelines please.

Issue #2080, opened 2016-04-04
Typeahead: call function on selection


spencerwi:
When a value is selected from a typeahead, it would be nice to be able to call a function (e.g. `on-change="ctrl.fetchData($selection)"` distinct from a simple ng-change function (which would trigger on every keypress).

For example, I have a project where the typeahead is used as a entry/search field across product codes. I want to allow people to manually enter a product code, but when a value is selected from a set of "known" product codes that populates the typeahead's list, I want to be able to fetch more details about the "known" product.

In a single page, there may be anywhere from 1 to 50 of these product code fields, so to go with an event-listening approach would mean that not only would I have to become dependent on injected `$scope` (which harms migration-readiness to Angular2 in my app and [diverges from Angular best practices](https://github.com/johnpapa/angular-styleguide/blob/master/a1/README.md#style-y030)), but I also would have to write quite a bit of boilerplate code to identify _which_ typeahead triggered the event and thus _which_ backing model should be updated with _what_ new value.

DaleWebb:
So when a typeahead option is selected, you want `ctrl.fetchData($selection)` to be fired?

Would a event be sufficient to satisfy the usecase?  So rather than putting your function call in the attribute, you can listen to an event fired by the typeahead and you would do: 

```javascript
...
$scope.$on('typeahead.selected', $scope.fetchData);
...
```

spencerwi:
In a single page, there may be anywhere from 1 to 50 of these product code fields, so to go with an event-listening approach would mean that not only would I have to become dependent on injected $scope (which harms migration-readiness to Angular2 in my app and diverges from Angular best practices), but I also would have to write quite a bit of boilerplate code to identify which typeahead triggered the event and thus which backing model should be updated with what new value.

DaleWebb:
Sorry, I should learn to read the whole thing...

You can attach the `data-prefix-event=""` attribute and give each field a specific name.

spencerwi:
Yeah, I assumed that was possible (otherwise the boilerplate code wouldn't be annoying boilerplate but would instead be giant horrible tightly-coupled DOM-querying code :stuck_out_tongue: ).

Still means that instead of adding a single attribute in my DOM that calls a controller function, I need to inject `$scope`, set up a listener on it, build a giant switch-case (or some other prefix-to-context-data mapper), _then_ call the function I was going to write in the first place.

[...]

Issue #1940, opened 2015-11-03
bs-select Add action on select


vapits:
That would be a great feature. Trigger an action (Ex a function) when an option is selected (well something like ng-click but in options).

vmlf01:
Maybe you can use the '.select' event triggered when an item is selected. Have a look at
http://plnkr.co/edit/8IaJEu5cnm2U1kvQdPGT?p=preview

vapits:
@vmlf01 thx, i already did that, i opened the issue so maybe that will become a future feature :) ;)

PR #2082, opened 2016-04-05
feat(callbacks): add function callbacks for events via attributes on directives

DaleWebb:
To perform an action on an event that a component will trigger (e.g. show a modal)
you have to listen to the event on the scope.
Now, you can attach the function to be called via attributes (`bs-on-*`).  This is used
for all events that are also emitted through the scope.

closes: #2080 #1940

### DIFF
--- a/src/alert/alert.js
+++ b/src/alert/alert.js
@@ -82,6 +82,14 @@ angular.module('mgcrea.ngStrap.alert', ['mgcrea.ngStrap.modal'])
           if (angular.isDefined(attr[key]) && falseValueRegExp.test(attr[key])) options[key] = false;
         });
 
+        // bind functions from the attrs to the show and hide events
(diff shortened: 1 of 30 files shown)


DaleWebb:
I added attributes: `bs-on-show`, `bs-on-before-show`, `bs-on-before-hide`, `bs-on-select`.
I went for the prefix of `bs-on` to avoid name clashes with other directive names and general HTML attributes.

mgcrea:
Great job! Landed in ce50472.

lsimmons2:
Hi, I'm using `bs-on-select='myFunction()'` for a typeahead which is repeated with `ng-repeat`. Instead of `myFunction()` being fired when I select something from any of the typeaheads, it is fired 8 times (the number of typeaheads in `ng-repeat`) when the page starts.

Any idea why this is happening?

Thanks!

Issue #1857, opened 2015-09-05
bs-select tab key causes error if no value is selected


jukkasi:
This can easily be reproduced with select demo

1. http://mgcrea.github.io/angular-strap/#/selects
2. Click "Choose among the following" to get focus
3. Press "Tab" key on keyboard
4. Error Uncaught TypeError: Cannot read property 'value' of undefined in console

fayzaan:
I have a fix for this, but how do I push it?

Trial-In-Error:
Hey! This was never merged & I'm still encountering this issue, both on the demo page (http://mgcrea.github.io/angular-strap) and in my own project. Is there anything I can do to help?

DaleWebb:
You could fork your own version for now and apply the simple change like the commit referenced above.

Then I'd welcome a pull request with the change :+1:

Trial-In-Error:
Hey! Thanks for the pull request; I'd been meaning to get to it, but been busy. Sorry. :(

PR #2084, opened 2016-04-07
fix(select): don't attempt to select when tab is pressed and no item is selected

DaleWebb:
Currently, when tab is pressed to get to a new input from `bs-select`, the `bs-select` will
close and fire `$select.select()`.  If an item isn't select, this will be `undefined` and throw an `Error`.
Now, there is a check that the index of the item to be selected isn't undefined or outside
the bounds of the `$matches`.

closes: #1857


mgcrea:
Thanks! Landed in 73dc51a.

PR #2085, opened 2016-04-08
feat(select): allow single-selection bs-select to be toggled

### LINKED ISSUE #1756: Feature Request: Allow unsetting of option in bsSelect
I have a bsSelect restricted to single selections but the fields is optional. I would like the ability to unselect by clicking on an item in the dropdown that is already selected.

DaleWebb:
Currently, when a `bs-select` is allowed one selection, you cannot
deselect your chosen item.

Now, you can provide an option where if you click the selected item, it will now deselect.

closes #1756


mgcrea:
Thanks! Landed in ceab80f.

PR #2087, opened 2016-04-08
fix(modal): do not decrement count if hide has been cancelled

just-boris:
Now you can cancel modal hide from event `modal.hide.before`, but backdropCount will be decremented, even if it is not removed.

After this fix `backdropCount` will change only if backdrop is actually removed.


mgcrea:
Thanks! Landed in c6a9072.

Issue #2086, opened 2016-04-08
Up and down keys not working as expected in timepicker in version 2.3.7


JaiL81:
Hello,

I am using timepicker and I have seen an strange behavior in tumepicker directive when using up/down arrow keys. My expectation is that the up/down arrow key works the same way the arrow up/down button works, this is adding/subtracting hours/minutes. 

However they work the other way around, up arrow key subtracts and down arrow key adds. I have checked the code and yes, that's the case and it looks weird for me, does it have any explanation? and is it configurable in some way?

I am using AngularJS 1.4.9 and angular-strap 2.3.7.

Thanks and best regards.
Jaime

DaleWebb:
Hi @JaiL81,

The up and down is to navigate the UI of the timepicker, rather than thinking about adding and subtracting.

This behaviour isn't configurable.

JaiL81:
Hi @DaleWebb ,

Thanks for your quick response. But don't you think that this behavior can be a little bin confusing for the user? If you have a timepicker with a length = 1 a and the arrow-behavior = picker, the user will see just one row in the UI tooltip and both actions of the up arrow and up key will behave in a different way and I as a user I would expect that both actions behave the same way, don't you think it would be a good feature that this would be at least configurable?

Thanks and best regards.
Jaime

DaleWebb:
Hmm, I'm not quite sure what you mean.  I think the arrows are for skipping sections of the numbers, whereas the keys are for selection.

JaiL81:
Hi @DaleWebb ,

Maybe with this plunkr I can explain myslef better: http://plnkr.co/edit/6NVx3OqpK3J4PcK5ce2S?p=preview

In this example extracted form the documentation you can see a timepicker with length = 1, and I think that a normal user would expect that in this case the up arrow button and the up arrow key behaves the same way, don't you think?

Thanks and best regards.
Jaime

DaleWebb:
Ah! I see.  I wasn't aware of the `data-arrow-behaviour="picker"` attribute. 

Yes, you're right, this behaviour should be reversed.

PR #2089, opened 2016-04-11
fix(timepicker): reverse behaviour of arrows in picker mode

DaleWebb:
currently when you specify the arrow behaviour to be "picker",
when you press "up" it will increment the number.  This contrasts
to the keyboard button "up" which will decrement.
Now, the behaviour is the same for both the keyboard and arrow, by
making the arrow "up" decrement instead.

closes: #2086


acurrieclark:
Does it not make more sense (in picker mode) for both the keyboard and arrow "up" actions to increment? When compared to the pager mode, I see why you have made the change as you have done, but in isolation it strikes me as more intuitive for up to increment and down to decrement.

DaleWebb:
If you think where the numbers would be if you could see them listed, I imagine it would be in ascending order.  Is your argument similar to touchpad scrolling, where some people prefer to reverse the scroll?

acurrieclark:
I know what you mean. But in my mind, pressing up would mean increasing,
rather than moving within a list. When you see the numbers in "pager" mode,
you see the list. Otherwise, the only number you see is the current time.
In that case "up" would really mean "plus".

DaleWebb:
Hmm, can you find a website that has implemented that behaviour?  I feel that the current behaviour is what the user would expect

acurrieclark:
It seems there are a few different take on this. But it would seem that the majority of timepickers I found have “up" as “plus” when in the mode we are describing. For example:

https://jdewit.github.io/bootstrap-timepicker/
http://ericjgagnon.github.io/wickedpicker/
http://keith-wood.name/timeEntry.html

This one does the opposite, but offers the option to change the default behaviour:

http://senthilraj.github.io/TimePicki/

I absolutely agree that if the times were listed, then a user would expect “up” to move up the list, hence decreasing the time. However, if the user is not presented with a list, I cannot help but feel that they would expect “up” to mean “plus”. On a site which used only this type of timepicker, a user would not even be used to seeing a list in the context.

Perhaps an option would be the best way forward?
All of this aside, the fact that the arrows now match the keyboard action is definitely a step in the right direction.

-- 
Alex Currie-Clark
07825346673

On 10 May 2016 at 23:37:20, Dale Alexander Webb (notifications@github.com) wrote:

Hmm, can you find a website that has implemented that behaviour? I feel that the current behaviour is what the user would expect

mgcrea:
Agree that this should be an option.

mgcrea:
@DaleWebb is this mergeable?

DaleWebb:
Yeah

PR #2093, opened 2016-04-18
Twitter Bootstrap was changed to Bootstrap

wuliupo:
Update Bootstrap website


mgcrea:
Thanks! Landed in 7ff462d.

PR #2097, opened 2016-04-27
Tooltip option to disable button mousedown preventDefault & stopPropagation

aronduby:
The `stopPropagation` call within tooltip button mousedown was interfering with other programming, so I added the ability to disable those


mgcrea:
Thanks! Landed in fa3fded.

PR #2098, opened 2016-05-03
Fix issue of leading space when title is not given in $alert

nnattawat:
I have faced the leading space in `$alert` when there is no `title` given and I found in the source that it's commented as `TODO`.

This PR is to fix that `TODO` issue.


mgcrea:
Thanks! Landed in 618367c.

PR #2115, opened 2016-05-26
fix(affix): various fixes (width, resize, bouncing, ngRoutes)

CyborgMaster:
* Allow the ability to turn off setting the width in the style using a property
* When calculating the offset, also remove any element style for top (right now
  only position is removed).  This is necessary in order to correctly
  recalculate offsets when the page resizes.
* Fix bouncing when it was affixed to the bottom.  Right now the _unpin property
  will cause a flicker between middle and bottom every other pixel scrolled.
* Add timeout to intializing the affix element.  This allows the page to render
  before calculating offsets and allows it to work when transcluded, as occures
  when using ngRoute.

This is a refinement of #1555 and also closes #1555.

### DIFF
--- a/src/affix/affix.js
+++ b/src/affix/affix.js
@@ -6,7 +6,8 @@ angular.module('mgcrea.ngStrap.affix', ['mgcrea.ngStrap.helpers.dimensions', 'mg
 
     var defaults = this.defaults = {
       offsetTop: 'auto',
-      inlineStyles: true
+      inlineStyles: true,
+      setWidth: true
     };
 
     this.$get = function ($window, debounce, dimensions) {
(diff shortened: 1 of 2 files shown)


CyborgMaster:
I'm not sure why the build failed.  The check that runs on a specific version of node passed just fine.  The one that was on "stable" died.  It looks like it tried to compile node and failed.  I don't think this is a problem with my PR.

DaleWebb:
Yeah it looks like one of the node modules failed to install during `npm install`

mgcrea:
@CyborgMaster any chance you could squash your commits into a single one that I can cherry pick easily? Thanks!

CyborgMaster:
Sure.  I'll take care of that this afternoon.

CyborgMaster:
K.  They've been squashed.  Let me know if you need anything else.

mgcrea:
Thanks! Landed in 765f7ac.

madflow:
Would it be possible to release v2.3.10 with this commit? The latest release is just 9 days old - but this would be great!

CyborgMaster:
Agreed.  I would love to go back to using the official release, but it didn't make it into 2.3.9.  I would even be fine using it from master, but it's isn't in the `dist` folder in master, and for some reason I can't use it from `src`.  Can we get a new release?

jwhitmarsh:
Ditto!

PR #2135, opened 2016-06-28
Document bs- prefix requirement for defining on* functions in attributes

### LINKED ISSUE #2133: Documentation for bs-on-* attributes not clear
I love the addition of the on-\* attributes added in #2082! :)

Unfortunately, when trying to use it, it took me a good bit of debugging all the way down to the source code to figure out that I had to append "bs-" to the beginning because the docs just say onShow, onHide, etc.  I think it would be clearer to put them down as bsOnShow, etc. in the docs.

Or was the idea to have the "bs-" part optional?  Because if so, it currently does not work without the "bs-" part.

I'm happy to make a PR if there's no objections.

SteveShaffer:
Document bs- prefix requirement for defining on\* functions in attributes

Fixes #2133

### DIFF
--- a/src/alert/docs/alert.demo.html
+++ b/src/alert/docs/alert.demo.html
@@ -183,7 +183,7 @@ so remove the option then. Just looks bad to have an option that doesn't work
           <td>function</td>
           <td></td>
           <td>
-            <p>If provided, this function will be invoked after the alert is shown.</p>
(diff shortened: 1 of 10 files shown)


saveman71:
Sad that is was never merged :(

I struggled 20mns to find that, after reading the code, thanks @SteveShaffer for posting the issue, first hit https://www.google.fr/search?q=bsOnShow

Issue #2150, opened 2016-08-05
Resolve/Locals copied BUG


knicefire:
bsCompiler uses angular.copy which causes issue on passing resolved values.
Resolved values therefore looses bindings.

``` javascript
var resolve = angular.copy(options.resolve || {});
var locals = angular.copy(options.locals || {});
```

I see no reasons of using copy method here.

PR #2151, opened 2016-08-05
fix($bsCompiler): No need of using copy method in compiler

knicefire:
Prevent loosing bindings in resolved values

related to bug #2150


DaleWebb:
Is there a test you can write for the original implementation to show that bindings are lost, then pass the test with your fix?

madflow:
Hi,

just a note. The $bsCompiler service was ported on July 12th 2015 from angular material:

https://github.com/mgcrea/angular-strap/commit/b272203f7704d2f66873d974beb20e4ac9fd4e7f

On July 14th 2015 - the material project removed the copying of the "resolves" and "locals":

https://github.com/angular/material/commit/a914501361f31c2beb5f335b395a5f9f45cf36c3

The discussion is here: https://github.com/angular/material/issues/3651

I would argue that it is save to merge this (@knicefire without the dist folder please, releases should not be made in PRs) when the tests pass. This feels more like a vendor update that is good to merge if there is no regression introduced.

Issue #2164, opened 2016-08-20
bsCheckbox, bsCheckboxGroup and bsRadioGroup do not update active class correctly when used with ngModelOptions debouncing


elkami12:
First thanks for this excellent work !

The problem is for bsCheckbox, bsCheckboxGroup and bsRadioGroup.
If these directives are used inside a form with an ng-model-options used to debounce default events, the active class of buttons are not updated correctly.
Here is a fiddle that shows the problem : https://jsfiddle.net/elkami/pkj3x9xr/

The question is : why directives are not using ngModelControler `$viewValue` instead of `$modelValue` to test if input is active ?

For example, incorrect line of code for bsRadioGroup :
``` javascript
 var isActive = angular.equals(controller.$modelValue, value);
```

Corrected one :
``` javascript
 var isActive = angular.equals(controller.$viewValue, value);
```

Remark: bsCheckbox with true-value attribute work well...
I'm going to make a pull request for that...

PR #2165, opened 2016-08-20
fix for bsCheckbox and bsRadio: use $viewValue instead of $modelValue for setting isActive state

elkami12:
use ngModelController `$viewValue` and not `$modelValue` because the `isActive` state (and so active class of buttons) should be synchronized with viewValue.

For example, if a ngModelOptions is used with debouncing, `$modelValue` still equals the previous value on `$render` execution

That's a fix for #2164. Here is a fiddle using patch : https://jsfiddle.net/elkami/pkj3x9xr/

Thanks for review


mgcrea:
Thanks for the PR!

PR #1788, opened 2015-07-20
Feature(modal): Add configurable z-index

palortoff:
Make Z-Index of modal and backdrop configurable to avoid interference with other elements

PR #2167, opened 2016-09-01
feat(modal): configurable z-index

madflow:
This PR introduces a configurable z-index for modals. Replaces the stale and outdated #1788.

The modal z-index value defaults to 1050 - the backdrop value is decremented by 10.


mgcrea:
Thanks!

Issue #1580, opened 2015-03-17
Select and touch events on IOS


kenborge:
There seems to be a bug with the Select implementation on IOS (tested version 8). If you click right on the text inside the dropdown menu nothing happens, the element isn't selected. But if you click outside the text then the element gets selected. 

This happens with the select box on the angular demo page too if viewed with an iphone

cookednoodles:
I am seeing the same thing; this doesn't appear to be happening on the chrome emulator so 
I presume it's localized to ios

kenborge:
The reason for it not happening in the chrome emulator would be that the "isTouch" check that angular-strap does is still false if when you emulate touch events in the browser. 

I tracked the problem down to a handler that is registered on 'touchstart' and prevents and stops propagation of this. This causes the click event that happens after the touch event to never fire, so the element is never clicked. There is code that tries to handle this:
````
if(isTouch) {
  var targetEl = angular.element(evt.target);
  targetEl.triggerHandler('click');
}
```
But triggerHandler does not propagate the event so if text inside the element is clicked the parent element with the onClick is not triggered

kenborge:
There does not seem to be a reason to stop propagation of the touchstart event on select box, so my current solution is to just register the handler on onMouseDown and not on touchstart

kenborge:
Didn't turn out to be as simple as stopping to listen for touchstart, ended up breaking clicking in some cases

benbenwilde:
There are some poor decisions happening here.
Why would you ever trigger a `click` on `mousedown`, or `touchstart`?

You should `click` on `touchend`.
If `touchmove` happens after `touchstart`, you should not `click` on `touchend`.
You should not stop propagation on `touchstart`.
You also should not stop propagation on `mousedown`.
You also should not close the select box when the input loses focus. You should close the select box when both the input AND the dropdown lose focus. You should additionally close it after clicking inside the dropdown if it is not a multiselect.

darkobits:
:+1: 

This directive is completely hosed on iOS.

antonvp:
Any updates on the issue?

PR #2180, opened 2016-10-05
fix($select): trigger click on closest anchor tag

### LINKED ISSUE #2047: Can't click item with icon on mobile using the Select control.
If have an icon or insert additional HTML on a select item you cannot click on that item.  There is a fork in the code for a touch event that expects the element to be passed in be the anchor tag, however this is not the case if you have an icon.  This can be duplicated with the docs, if you open the http://mgcrea.github.io/angular-strap/#/selects demo on a mobile device you can reproduce this.  All that needs to happen is that the DOM needs to be walked until it finds the first anchor tag.  I think this should work on the $select.$onMouseDown event.  I'm sure there is a more efficient way to write this, but this is my stab at it.  I have this as a work around, but I would love to see this bug fixed.

              $select.$onMouseDown = function (evt) {
                  // Prevent blur on mousedown on .dropdown-menu
                  evt.preventDefault();
                  evt.stopPropagation();
                  // Emulate click for mobile devices
                  if (isTouch) {
                      var targetEl = angular.element(evt.target),
                          anchor;

                      if (evt.target.nodeName == 'A') {
                          anchor = targetEl;
                      }
                      else {
                          var anchors = targetEl.parents().toArray().filter(function (obj) {
                              return obj.nodeName == 'A';
                          });

                          if (anchors.length > 0)
                              anchor = anchors[0];
                      }

                      if (anchor != null) {
                          angular.element(anchor).triggerHandler('click');
                      }
                  }
              };

matthias-g:
On touch devices a click event is triggered on touchstart on an entry of the select list.
Trigger this emulated click not on the event.target but on the closest parent that is an anchor tag, where the click event triggers the select action.

Fix #2047 and #1580

### DIFF
--- a/src/select/select.js
+++ b/src/select/select.js
@@ -192,7 +192,23 @@ angular.module('mgcrea.ngStrap.select', ['mgcrea.ngStrap.tooltip', 'mgcrea.ngStr
           // Emulate click for mobile devices
           if (isTouch) {
             var targetEl = angular.element(evt.target);
-            targetEl.triggerHandler('click');
+            var anchor;
+
+            if (evt.target.nodeName !== 'A') {
+              var anchorCandidate = targetEl.parent();
+              while (!anchor && anchorCandidate.length > 0) {
+                if (anchorCandidate[0].nodeName === 'A') {
+                  anchor = anchorCandidate;
+                }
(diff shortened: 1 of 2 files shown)


mgcrea:
Thanks!

PR #2189, opened 2016-11-10
added index.js and module.export for compatibility with module bundlers

jverrone3:
This will add support for things like webpack and browserify.

## usage

```
angular.module('app.module', [require('angular-strap')]);
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,5 +1,6 @@
 // support for browserify and webpack
 
+requite('./dist/angular-strap.compat.min');
 require('./dist/angular-strap.min');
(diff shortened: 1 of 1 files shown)


madflow:
This is already addressed in https://github.com/mgcrea/angular-strap/pull/2152. Also - I believe either the "compat" build or the "normal" build should be included. Not both. I do not think webpack or browserify can tree-shake that off right now  which one you actually using.

mgcrea:
Fixed by #2152.

PR #2195, opened 2016-12-14
fix(tests), make tests pass even in december

madflow:
* Make datepicker tests pass in December ;) 
* `datepicker.spec.js:308:29 - Expected '1/1/17' to be '13/1/16'.`


mgcrea:
Nice :)

Issue #2179, opened 2016-10-04
Popover does not appear on demo page when using $popover service


Jokero:
You can check it here http://mgcrea.github.io/angular-strap/#/popovers-examples

Chrome 53.0.2785.116

```
TypeError: e.preventDefault is not a function
    at F.toggle (angular-strap.js:4440)
    at angular.js:15961
    at n.$eval (angular.js:17229)
    at n.$digest (angular.js:17045)
    at n.$apply (angular.js:17337)
    at HTMLButtonElement.<anonymous> (angular.js:25023)
    at Qf (angular.js:3456)
    at HTMLButtonElement.d (angular.js:3444)
```

LennyLip:
+1 no errors, last Firefox, Chrome, Opera, Explorer 11

cycold:
Yeah, there is no `preventDefault method` in  `evt`  see below pic 
```
if (evt) {
     evt.preventDefault && evt.preventDefault();
 }
```
![image](https://cloud.githubusercontent.com/assets/2446601/22924244/5d42fbc2-f2df-11e6-85f3-610bdf2cb2bd.png)

xeroxoid:
Hi all, as I do not see the pull request being merged any time soon, a workaround is to use:

```myPopover.toggle();```

i.e. directly instead of after the promise as the example

```myPopover.$promise.then(myPopover.toggle);```

PR #2206, opened 2017-02-14
fix(tooltip) poor man's fix for #2179

madflow:
This is a quick and hacky fix for #2179.


DaleWebb:
What are the side-effects for not executing `preventDefault()`?

This might be covering up a bigger problem, where `evt` is not the correct thing to be passed in (looks like it's the element rather than event)

madflow:
@DaleWebb When the popover is triggered via markup - `evt` it is an MouseEvent object. When triggered manually via `$toggle` - it is an object with the element and `locals` etc.

How would one determine the "event" when triggered manually?  This could be anything I guess.

PR #2221, opened 2017-03-27
fix(datepicker): hide when autoclose set to true

mikejr83:
Setting "autoclose" to true and "trigger" to "click" closes the datepicker dropdown when mouse event
happens on the body.

Fix: The datepicker extends tooltip. Tooltip looks for 'autoClose' whereas
datepicker expects 'autoclose'. The directive passes through 'autoclose' to
'autoClose'.


mikejr83:
I think I misinterpreted how the autoclose was supposed to work in the click trigger. This appears to cause regressions in other triggers.

PR #2228, opened 2017-04-11
Option to allow datepicker for readonly elements

asednev:
Bug fix [#946](https://github.com/mgcrea/angular-strap/issues/946) removed calendar for read-only elements which can be useful behavior. This PR allows to explicitly enable the behavior that was removed in "datePicker and timePicker show calendar when ng-readonly="true" #946"


asednev:
@DaleWebb @mgcrea any ETA for merging this PR?

PR #2229, opened 2017-04-11
feat(modal): 508 support in modal

mikejr83:
508 and screen-reader support in modal.


mikejr83:
FYI, unit test cases are included in the feature commit.

PR #2233, opened 2017-04-25
fix(bsSelect): Use $viewChangeListeners instead of $watch

### LINKED ISSUE #1966: bsSelect does not work with a Breeze entity in ngModel
Breeze entities have circular references.  Setting the ngModel to a Breeze entity will cause the stack to overflow since the bsSelect is doing a deep comparison on the watch function.

I was able to make it work by removing the 'true' parameter on the $watch method on ngModel on line 2642:

```
    scope.$watch(attr.ngModel, function(newValue, oldValue) {
      select.$updateActiveIndex();
      controller.$render();
    });
```

I don't know if this is always desirable so maybe this can be set on the directive as a optional parameter?

fbjon:
This fixes the case where ngModelOptions specify getterSetter: true. In that
case watching the ngModel attribute does nothing useful, because it will point to
a function rather than a value.

- The directive no longer deep-watches the model (this is default ngModel behaviour)
- No deep-watching should also fix #1966

### DIFF
--- a/src/select/select.js
+++ b/src/select/select.js
@@ -366,17 +366,17 @@ angular.module('mgcrea.ngStrap.select', ['mgcrea.ngStrap.tooltip', 'mgcrea.ngStr
         }, true);
 
         // Watch model for changes
-        scope.$watch(attr.ngModel, function (newValue, oldValue) {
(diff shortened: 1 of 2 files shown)


fbjon:
This fixes #2235.

PR #2238, opened 2017-05-08
Use the document instead of the body to determine the height of the page

CyborgMaster:
In my project, due to fixed height headers and sidebars, the children of the body element are all absolutely, or fixed positioned.  This causes the body element to have a height of 0px, and therefore the height calcs in affix to work incorrectly.  This is fixed if we use the height of the document instead of the body.


CyborgMaster:
I really think this is a good improvement, I would like to get it merged.  Can we take another look?

PR #2262, opened 2017-10-25
fix(datepicker): unable to select the same date as minDate

prashanth2692:
Due to passing string with 'Z' into Date constructor the date value is being offset by browser's timezone which is causing the the datepicker to disable the day which is passed as minDate. Effectively not allowing to select the date which is set as minDate. This will be reproducible for user on East of UTC.

To fix the issue, checking for 'Z' in the date string and excluding it to get the Date object with the expected date.


prashanth2692:
@mgcrea , can you please have a look at this PR.

